PR #1, opened 2016-10-11
Support any StaticVector indexing an Extrapolation

andyferris:
`Tuple()` is part of the `StaticArray` interface. Allows for `MVector` and custom (e.g. `Point`) types.


timholy:
Shoot, I didn't notice this earlier. Seems fine, thanks!

PR #3, opened 2016-10-13
update REQUIRE

Evizero:
The tests still don't run locally for me because `recenter` is not defined. I am guessing WIP?

PR #4, opened 2017-02-12
move over restrict from Images

Evizero:
I am working on moving over `restrict` and `resize` over here from Images.jl

A few question at this point

- What do we do with the `restrict` methods for `AxisArray` and `ImageMetaAxis` (https://github.com/JuliaImages/Images.jl/blob/master/src/algorithms.jl#L516-L531). Do we move them here as well and introduce a dependency on the corresponding packages?

- ~~where does `recenter` in the tests come from, was this some unmerged change somewhere?~~

### DIFF
--- a/src/resizing.jl
+++ b/src/resizing.jl
@@ -127,3 +127,32 @@ end
 
 restrict_size(len::Integer) = isodd(len) ? (len+1)>>1 : (len>>1)+1
 
+## imresize
+function imresize!(resized, original::AbstractMatrix)
+    scale1 = (size(original,1)-1)/(size(resized,1)-0.999f0)
+    scale2 = (size(original,2)-1)/(size(resized,2)-0.999f0)
+    for jr = 0:size(resized,2)-1
+        jo = scale2*jr
+        ijo = trunc(Int, jo)
+        fjo = jo - oftype(jo, ijo)
+        @inbounds for ir = 0:size(resized,1)-1
+            io = scale1*ir
+            iio = trunc(Int, io)
+            fio = io - oftype(io, iio)
+            tmp = (1-fio)*((1-fjo)*original[iio+1,ijo+1] +
+                           fjo*original[iio+1,ijo+2]) +
+                  + fio*((1-fjo)*original[iio+2,ijo+1] +
+                         fjo*original[iio+2,ijo+2])
+            resized[ir+1,jr+1] = convertsafely(eltype(resized), tmp)
+        end
+    end
+    resized
+end
+
+imresize(original, new_size) = size(original) == new_size ? copy!(similar(original), original) : imresize!(similar(original, new_size), original)
+
+convertsafely{T<:AbstractFloat}(::Type{T}, val) = convert(T, val)
+convertsafely{T<:Integer}(::Type{T}, val::Integer) = convert(T, val)
+convertsafely{T<:Integer}(::Type{T}, val::AbstractFloat) = round(T, val)
+convertsafely{T}(::Type{T}, val) = convert(T, val)
+
--- a/src/resizing.jl
+++ b/src/resizing.jl
@@ -1,0 +1,129 @@
+"""
+`imgr = restrict(img[, region])` performs two-fold reduction in size
+along the dimensions listed in `region`, or all spatial coordinates if
+`region` is not specified.  It anti-aliases the image as it goes, so
+is better than a naive summation over 2x2 blocks.
+"""
+restrict(img::AbstractArray, ::Tuple{}) = img
+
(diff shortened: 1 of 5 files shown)


Evizero:
I just discovered I had the exact same question about `recenter` before at #3

timholy:
I think it would be fine to put `AxisArray` methods in `ImageAxes.jl` and `ImageMetaAxis` methods in `ImageMeta`.

src/resizing.jl
```diff
@@ -0,0 +1,158 @@
...
+    if size(A, dim) <= 2
+        return A
+    end
+    newsz = ntuple(i->i==dim?restrict_size(size(A,dim)):size(A,i), Val{N})
+    out = Array{typeof(A[1]/4+A[2]/2),N}(newsz)
+    restrict!(out, A, dim)
+    out
+end
+
+# out should have efficient linear indexing
```
timholy: Ugh, I'd forgotten that `restrict` still used the old `Base.Cartesian` rather than `CartesianIndex`. Keep it like it is now, but someday I should change this.

src/resizing.jl
```diff
@@ -0,0 +1,158 @@
...
+                end
+            end
+        end
+    end
+end
+
+restrict_size(len::Integer) = isodd(len) ? (len+1)>>1 : (len>>1)+1
+
+## imresize
+function imresize!(resized, original::AbstractMatrix)
```
timholy: Eventually needs to be made multidimensional. (Again, not your responsibility unless you want to tackle that.) Really we should use Interpolations, which will give us much more flexibility with the interpolation scheme and boundary conditions.

Evizero:
> I think it would be fine to put AxisArray methods in ImageAxes.jl and ImageMetaAxis methods in ImageMeta.

So these packages would then depend on `ImageTransformations`?

timholy:
Hmm. Right, not a good idea. Better to just leave them in Images.

Evizero:
> Again, not your responsibility unless you want to tackle that.

Both suggestions sounds like something I would like to attempt. These seems like good introduction task to get my feet wet with this package.

That said I think we should first complete to move. I have seen over and over again that it is hard for me to estimate when I'll be swamped at work again, which caused half-finished branches in the past that sat idle for a while. So I'd like to get into the habit of finishing tasks one by one.

Evizero:
> Hmm. Right, not a good idea. Better to just leave them in Images.

One thing that works pretty well over at JuliaML is the following: 
- We define some common function **names** in the base package (like `function restrict end`), and
- then the package that is responsible for that functionality (i.e. ImageTransformations) implements the core workhorse methods for it. 
- Other packages that need to **decorate** that functionality do so themselves (without knowing the responsible package) by extending the base package only and calling the signature that the responsible package provides. 

This seems like such a use-case. thoughts?

src/resizing.jl
Evizero: A related question: Was this method implemented before generated functions existed, or was there a reason not to use them?

timholy: Yes. Images was a big part of the motivation for introducing generated functions.

Evizero:
New question / suggestion (sorry for making so much noise, but I can't make these kind of decisions myself)

There are two function `restrict` and `_restrict`. The first performs on multiple dimensions and the later just at one. I am guessing the idea was to hide the later from the user to discourage him/her from using it. (https://github.com/JuliaImages/Images.jl/blob/master/src/algorithms.jl#L550-L558)

For axis arrays however the corresponsing one-dimension function that decorates `_restrict` is called `restrict` (https://github.com/JuliaImages/Images.jl/blob/master/src/algorithms.jl#L523-L526)

For consistency, should we change `_restrict` to be just another method signature for `restrict` (i.e. rename it)?

timholy:
Yes, it seems reasonable to make that a public method. Thanks for asking!

src/resizing.jl
```diff
@@ -0,0 +1,165 @@
+### restrict, for reducing the image size by 2-fold
+# This properly anti-aliases. The only "oddity" is that edges tend towards zero under
+# repeated iteration.
+# This implementation is considerably faster than the one in Grid,
+# because it traverses the array in storage order.
+if isdefined(:restrict)
```
timholy: Since Grid is a deprecated package, let's just get rid of this.

Evizero:
I think I got everything concerning `restrict` and `imresize`. The tests for `ImageTransformations` and the corresponding `Images` branch work locally. Anything else I could/should do?

right now the `restrict` methods for `AxisArray` and `ImageMeta` are still in `Images`, as you suggested

timholy:
Thanks! I'll register this.

PR #6, opened 2017-02-13
add lower versions to REQUIRE

Evizero:
- Colors and ColorVectorSpace I copied the values that Images requires (just in case).
- Interpolations I based it on a recent change you did that concerned BSplines.
- ImageCore seems reasonable

I am a bit lost what to look for at OffsetArrays and StaticArrays. Since Images.jl doesn't list any lower version I just left them as is

### DIFF
--- a/REQUIRE
+++ b/REQUIRE
@@ -1,8 +1,8 @@
 julia 0.5
-ImageCore
+ImageCore 0.1.0
 CoordinateTransformations
-Interpolations
+Interpolations 0.3.3
 OffsetArrays
 StaticArrays
-Colors
-ColorVectorSpace
+Colors 0.7.0
+ColorVectorSpace 0.2
--- a/REQUIRE
+++ b/REQUIRE
@@ -1,7 +1,7 @@
 julia 0.5
-ImageCore 0.1.0
+ImageCore 0.1.2
 CoordinateTransformations
-Interpolations 0.3.3
+Interpolations 0.3.7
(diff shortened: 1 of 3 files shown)


Evizero:
Tests wont run until `recenter` is available at CoordinateTransforms, which is not yet tagged. I made a request there: https://github.com/FugroRoames/CoordinateTransformations.jl/issues/24

Evizero:
We could comment out the `warp` related tests for now and focus on the `restrict` and `imresize` move (?)

timholy:
I'd be fine with you commenting those out, if that's what you prefer.

Evizero:
How should we go about 0.6 testing? The error thrown in appveyor on nightly seems to be an upstream one in Interpolations, which in turn looks like it was fixed on master but not yet tagged

timholy:
I'll tag a new Interpolations release.

REQUIRE
```diff
@@ -1,8 +1,8 @@
 julia 0.5
-ImageCore
+ImageCore 0.1.0
 CoordinateTransformations
```
andyferris: will be 0.3.3 for recenter

andyferris: Sorry v0.4.0 now. https://github.com/JuliaLang/METADATA.jl/pull/7964

Evizero:
I'll go with this for now. I'll reactivate and bump CoordinateTransformations later.

PR #8, opened 2017-02-27
Improve the coordinate mapping in imresize

timholy:
`imresize` has to decide how points in the resized image correspond to points in the original image. We've had behavior that can be viewed as somewhat inconsistent:
```julia
julia> using Images   # using the old algorithm before it moved to ImageTransformations

julia> import ImageTransformations   # what's currently on master, not this PR

julia> A = [1 0; 0 0]
2×2 Array{Int64,2}:
 1  0
 0  0

julia> Images.imresize(A, (1,1))
1×1 Array{Int64,2}:
 1

julia> ImageTransformations.imresize(A, (1,1))
1×1 Array{Int64,2}:
 0

julia> A = [0 1; 0 0]
2×2 Array{Int64,2}:
 0  1
 0  0

julia> Images.imresize(A, (1,1))
1×1 Array{Int64,2}:
 0

julia> ImageTransformations.imresize(A, (1,1))
1×1 Array{Int64,2}:
 0

julia> A = [0 0; 0 1]
2×2 Array{Int64,2}:
 0  0
 0  1

julia> Images.imresize(A, (1,1))
1×1 Array{Int64,2}:
 0

julia> ImageTransformations.imresize(A, (1,1))
1×1 Array{Int64,2}:
 1
```
With this PR, one gets 0.25 for all of these operations.

For other sizes it will not be quite so invariant, but with this version we at least have a slightly more "principled" choice for how we map coordinates, see https://github.com/JuliaImages/ImageTransformations.jl/blob/d0b802e313240fe5443bcc6d4ed33b9dae22d32a/src/resizing.jl#L160-L169.

### DIFF
--- a/src/ImageTransformations.jl
+++ b/src/ImageTransformations.jl
@@ -1,7 +1,7 @@
 __precompile__()
 module ImageTransformations
 
-using CoordinateTransformations, Interpolations, OffsetArrays, StaticArrays, Colors, ColorVectorSpace, ImageCore
+using CoordinateTransformations, Interpolations, OffsetArrays, StaticArrays, FixedPointNumbers, Colors, ColorVectorSpace, ImageCore
 
 import Base: start, next, done, eltype, iteratorsize
 using Base: tail, Cartesian
@@ -93,4 +93,3 @@ center{T,N}(img::AbstractArray{T,N}) = SVector{N}(map(_center, indices(img)))
 _center(ind::AbstractUnitRange) = (first(ind)+last(ind))/2
 
 end # module
-
--- a/src/resizing.jl
+++ b/src/resizing.jl
@@ -162,11 +162,37 @@ function imresize!{T,S,N}(resized::AbstractArray{T,N}, original::AbstractArray{S
 end
 
 function imresize!{T,S,N}(resized::AbstractArray{T,N}, original::AbstractInterpolation{S,N})
-    sf = map(/, size(original), size(resized))
-    @inbounds for I in CartesianRange(size(resized))
-        I_o = map(*, I.I, sf)
-        resized[I] = original[I_o...]
+    # Define the equivalent of an affine transformation for mapping
+    # locations in `resized` to the corresponding position in
+    # `original`. We take the viewpoint that a pixel at `i, j` is a
+    # sensor that *integrates* the intensity over an area spanning
+    # `i±0.5, j±0.5` (this is a good model of how a camera pixel
+    # actually works). We then map the *outer corners* of the two
+    # images to each other, i.e., in typical cases
+    #     (0.5, 0.5) -> (0.5, 0.5)  (outer corner, top left)
+    #     size(resized)+0.5 -> size(original)+0.5  (outer corner, lower right)
+    # This ensures that both images cover exactly the same area.
+    Ro, Rr = CartesianRange(indices(original)), CartesianRange(indices(resized))
+    sf = map(/, (last(Ro)-first(Ro)+1).I, (last(Rr)-first(Rr)+1).I) # +1 for outer corners
+    offset = map((io,ir,s)->io - 0.5 - s*(ir-0.5), first(Ro).I, first(Rr).I, sf)
+    if all(x->x >= 1, sf)
+        @inbounds for I in Rr
+            I_o = map3((i,s,off)->s*i+off, I.I, sf, offset)
+            resized[I] = original[I_o...]
+        end
+    else
+        @inbounds for I in Rr
+            I_o = clampR(map3((i,s,off)->s*i+off, I.I, sf, offset), Ro)
+            resized[I] = original[I_o...]
+        end
     end
     resized
 end
 
+# map isn't optimized for 3 tuple-arguments, so do it here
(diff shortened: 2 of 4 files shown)


src/resizing.jl
```diff
@@ -137,11 +141,30 @@ function imresize{T,N}(original::AbstractArray{T,N}, short_size::NTuple)
...
 function imresize{T,N}(original::AbstractArray{T,N}, new_size::NTuple{N})
+    # To choose the output type, rather than forcing everything to
+    # Float64 by multiplying by 1.0, we exploit the fact that the scale
+    # changes correspond to integer ratios.  We mimic ratio arithmetic
+    # without actually using Rational (which risks promoting to a
+    # Rational type, too slow for image processing).
+    Tnew = typeof((first(original) * 1) / 1)
     if size(original) == new_size
-        copy!(similar(original), original)
+        copy!(similar(original, Tnew), original)
```
Evizero: This is primarily done so `imresize` is typestable, right?

src/resizing.jl
```diff
@@ -137,11 +141,30 @@ function imresize{T,N}(original::AbstractArray{T,N}, short_size::NTuple)
...
+
+See also [`restrict`](@ref).
+"""
 function imresize{T,N}(original::AbstractArray{T,N}, new_size::NTuple{N})
+    # To choose the output type, rather than forcing everything to
+    # Float64 by multiplying by 1.0, we exploit the fact that the scale
+    # changes correspond to integer ratios.  We mimic ratio arithmetic
+    # without actually using Rational (which risks promoting to a
+    # Rational type, too slow for image processing).
+    Tnew = typeof((first(original) * 1) / 1)
```
Evizero: Should we consider treating Colorant arrays differently? 

```julia
julia> Gray{N0f8}(0.8) * 1 / 1
Gray{Float32}(0.8000001f0)
```

Intuitively I would expect an array of `Gray{N0f8}` to remain of that eltype after resizing.

src/resizing.jl
```diff
@@ -151,11 +174,37 @@ function imresize!{T,S,N}(resized::AbstractArray{T,N}, original::AbstractArray{S
...
+            I_o = clampR(map3((i,s,off)->s*i+off, I.I, sf, offset), Ro)
+            resized[I] = original[I_o...]
+        end
     end
     resized
 end
 
+# map isn't optimized for 3 tuple-arguments, so do it here
+@inline map3(f, a, b, c) = (f(a[1], b[1], c[1]), map3(f, tail(a), tail(b), tail(c))...)
+@inline map3(f, ::Tuple{}, ::Tuple{}, ::Tuple{}) = ()
```
Evizero: does this recursively inline? How is this fast? (just curious)

src/ImageTransformations.jl
```diff
@@ -93,4 +93,3 @@ center{T,N}(img::AbstractArray{T,N}) = SVector{N}(map(_center, indices(img)))
 _center(ind::AbstractUnitRange) = (first(ind)+last(ind))/2
 
 end # module
-
```
Evizero: :-(

src/resizing.jl
timholy: Hmm, yes, since those "mimic" floating-point (meaning, they round automatically) that might be better. Good catch.

src/resizing.jl
timholy: Correct.

src/resizing.jl
timholy: Yes, the `@inline` forces Julia's compiler to perform all these operations at compile time rather than run time. What this is basically doing is expressing the computation in a way that Julia's compiler can make sense of it: the compiler is extremely good at reasoning about tuples, even when the elements of the tuple are of different types, and consequently generates excellent code. Virtually all of Julia's built-in array operations (`getindex`, `setindex!`, `broadcast`, etc.) are coded in this style of programming. Without the `@inline` this would be a disaster, though.

src/ImageTransformations.jl
timholy: No worries, I'm *really* unconcerned about things like whitespace :smile:. It's only the fact that I have a setting in emacs to automatically get rid of it that this got caught. You're helping a lot!

src/resizing.jl
Evizero: nice! that is good to know, thanks

timholy:
CI seems stalled, let's see if reopening helps.

src/ImageTransformations.jl
```diff
@@ -1,7 +1,7 @@
 __precompile__()
 module ImageTransformations
 
-using CoordinateTransformations, Interpolations, OffsetArrays, StaticArrays, Colors, ColorVectorSpace, ImageCore
+using CoordinateTransformations, Interpolations, OffsetArrays, StaticArrays, FixedPointNumbers, Colors, ColorVectorSpace, ImageCore
```
Evizero: Tony will yell at us if we don't put `FixedPointNumbers` in `REQUIRE` if we use it directly

timholy: Thanks! Turns out I had a rebase-induced bug, so this has to do another round anyway.

PR #9, opened 2017-03-13
Add docstring for `warp`

timholy:
This should make `warp` more accessible. The functionality in the example works now; however, people will quickly discover limitations. TODO:
- [x] fix Interpolations.jl so it can work with non-1 based indices
- [ ] advertise `warp` in the docstring for `?Images`
- [ ] document `warp` at JuliaImages and update [this page](http://juliaimages.github.io/latest/indexing.html#Keeping-track-of-location-with-unconventional-indices-1).

### DIFF
--- a/src/warp.jl
+++ b/src/warp.jl
@@ -26,6 +26,67 @@ end
 
 warp{T}(img::AbstractInterpolation{T}, tform, fill=_default_fill(T)) = warp(extrapolate(img, fill), tform)
 
+"""
+    warp(img, tform) -> imgw
+
+Transform the coordinates of `img`, returning a new `imgw` satisfying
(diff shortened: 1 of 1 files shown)


Evizero:
👍 .

screen shot 2017-03-13 at 23 08 17

On a related note, that example made me spot a bug in ImageInTerminal

timholy:
I noticed the bug too; `restrict` needs fixing for non-1 indices, so this goes further than just ImagesInTerminals. I'll submit what I have so far.

Evizero:
My guess is this PR is pretty hopelessly in conflict with master now. Might be easier to just copy and paste the doc-string and make a brand new commit from current master

PR #10, opened 2017-03-14
make wrap preserve eltype of Colorant Arrays

Evizero:
It seems more intuitive to me to not promote to Float32 / Float64 automatically if fixed point numbers are used. At least it should be a conscious user decision. We could introduce a signature `wrap(::Type{NewElType}, ...)` to expose that option.

Thoughts?

### DIFF
--- a/src/ImageTransformations.jl
+++ b/src/ImageTransformations.jl
@@ -25,6 +25,12 @@ function warp(img::AbstractExtrapolation, tform)
     out = OffsetArray(Array{eltype(img)}(map(length, inds)), inds)
     warp!(out, img, tform)
 end
+function warp{TCol<:Colorant}(img::AbstractArray{TCol}, tform)
+    TNorm = eltype(TCol)
+    apad, pad = Interpolations.prefilter(TNorm, TCol, img, typeof(BSpline(Linear())), typeof(OnGrid()))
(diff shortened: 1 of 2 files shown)


timholy:
There's a lot of sense in this, but I'm also torn. In 0.5, we say "the array eltype wins" for broadcasting operations like `A .+ c`. But this is changing in 0.6 to produce whatever the corresponding scalar operation would yield. The other advantage I see of using `AbstractFloat` types is that `NaN` is a clearer marker for beyond-the-edge pixels than `zero` (the image might have 0 elements in it already).

However, I agree that it's not ideal now; while we provide a `warp!(out, img, tform)` version, computation of the output indices is not exported so it's hard to allocate `out` so that it covers the entire output region. I unequivocally like the idea of supplying the output eltype as an argument. I could also be persuaded to accept the rest, but I think that's worth reflecting on a bit more.

Evizero:
Good point. Well, to be fair all I really care about is that I have a simple option to preserve the Colortype, I don't at all need it to be the default behaviour. It just made sense to me intuitively. 
I agree that we should go with the general "Julian" way of doing things if we can

Regardless of this, it turns out I got some unexpected downtime right now so I am cleaning up the whole codebase a bit. I'll throw this PR out and open a new one later. I'll try to come up with a more conservative handling of this eltype issue.

PR #11, opened 2017-03-14
RFC: Code refactor

Evizero:
I am in the process of cleaning up the code a bit and writing test. This is not even close to merge ready, but I already made some changes that may be controversial so I just wanted to give the option of early feedback. 

One thing I did is move related code blocks into files, which granted may at this point not really be necessary. Yet it seems better to do it early, while the code-base is small enough and changes are easier to follow

### DIFF
--- a/src/ImageTransformations.jl
+++ b/src/ImageTransformations.jl
@@ -25,6 +25,12 @@ function warp(img::AbstractExtrapolation, tform)
     out = OffsetArray(Array{eltype(img)}(map(length, inds)), inds)
     warp!(out, img, tform)
 end
+function warp{TCol<:Colorant}(img::AbstractArray{TCol}, tform)
+    TNorm = eltype(TCol)
+    apad, pad = Interpolations.prefilter(TNorm, TCol, img, typeof(BSpline(Linear())), typeof(OnGrid()))
+    itp = Interpolations.BSplineInterpolation(TNorm, apad, BSpline(Linear()), OnGrid(), pad)
+    warp(itp, tform)
+end
 warp(img::AbstractArray, tform) = warp(interpolate(img, BSpline(Linear()), OnGrid()), tform)
 warp{T<:FloatLike}(img::AbstractInterpolation{T}, tform) = warp(extrapolate(img, convert(T, NaN)), tform)
 warp{T<:FloatColorant}(img::AbstractInterpolation{T}, tform) = warp(extrapolate(img, nan(T)), tform)
--- a/src/ImageTransformations.jl
+++ b/src/ImageTransformations.jl
@@ -1,99 +1,28 @@
 __precompile__()
 module ImageTransformations
 
-using CoordinateTransformations, Interpolations, OffsetArrays, StaticArrays, FixedPointNumbers, Colors, ColorVectorSpace, ImageCore
+using ImageCore
+using CoordinateTransformations
+using StaticArrays
+using Interpolations
+using OffsetArrays
+using FixedPointNumbers
+using Colors, ColorVectorSpace
+using Compat
 
-import Base: start, next, done, eltype, iteratorsize
+import Base: start, next, done, eltype, iteratorsize, size, length
 using Base: tail, Cartesian
-
-typealias FloatLike{T<:AbstractFloat} Union{T,Gray{T}}
-typealias FloatColorant{T<:AbstractFloat} Colorant{T}
 
 export
 
-    warp,
-    center,
     restrict,
-    imresize
+    imresize,
+    warp,
+    center
 
+include("autorange.jl")
 include("resizing.jl")
-
-@inline Base.getindex(A::AbstractExtrapolation, v::StaticVector) = A[convert(Tuple, v)...]
-
-function warp(img::AbstractExtrapolation, tform)
-    inds = autorange(img, tform)
-    out = OffsetArray(Array{eltype(img)}(map(length, inds)), inds)
-    warp!(out, img, tform)
-end
-function warp{TCol<:Colorant}(img::AbstractArray{TCol}, tform)
-    TNorm = eltype(TCol)
-    apad, pad = Interpolations.prefilter(TNorm, TCol, img, typeof(BSpline(Linear())), typeof(OnGrid()))
-    itp = Interpolations.BSplineInterpolation(TNorm, apad, BSpline(Linear()), OnGrid(), pad)
-    warp(itp, tform)
-end
-warp(img::AbstractArray, tform) = warp(interpolate(img, BSpline(Linear()), OnGrid()), tform)
-warp{T<:FloatLike}(img::AbstractInterpolation{T}, tform) = warp(extrapolate(img, convert(T, NaN)), tform)
-warp{T<:FloatColorant}(img::AbstractInterpolation{T}, tform) = warp(extrapolate(img, nan(T)), tform)
-warp{T}(img::AbstractInterpolation{T}, tform) = warp(extrapolate(img, zero(T)), tform)
-
-function warp!(out, img::AbstractExtrapolation, tform)
-    tinv = inv(tform)
-    for I in CartesianRange(indices(out))
-        out[I] = img[tinv(SVector(I))]
-    end
-    out
-end
-
-function autorange(img, tform)
-    R = CartesianRange(indices(img))
-    I = first(R)
-    x = tform(SVector(I))
-    mn = mx = x
-    for I in CornerIterator(R)
-        x = tform(SVector(I))
-        mn, mx = min(x, mn), max(x, mx)
-    end
-    _autorange(convert(Tuple, mn), convert(Tuple, mx))
-end
-
-@noinline _autorange(mn,mx) = map((a,b)->floor(Int,a):ceil(Int,b), mn, mx)
-
-
-## Iterate over the corner-indices of a rectangular region
-immutable CornerIterator{I<:CartesianIndex}
-    start::I
-    stop::I
-end
-CornerIterator{I<:CartesianIndex}(R::CartesianRange{I}) = CornerIterator{I}(R.start, R.stop)
-
-eltype{I}(::Type{CornerIterator{I}}) = I
-iteratorsize{I}(::Type{CornerIterator{I}}) = Base.HasShape()
-
-@inline function start{I<:CartesianIndex}(iter::CornerIterator{I})
-    if any(map(>, iter.start.I, iter.stop.I))
-        return iter.stop+1
-    end
-    iter.start
-end
-@inline function next{I<:CartesianIndex}(iter::CornerIterator{I}, state)
-    state, I(inc(state.I, iter.start.I, iter.stop.I))
-end
-# increment & carry
-@inline inc(::Tuple{}, ::Tuple{}, ::Tuple{}) = ()
-@inline inc(state::Tuple{Int}, start::Tuple{Int}, stop::Tuple{Int}) = (state[1]+(stop[1]-start[1]),)
-@inline function inc(state, start, stop)
-    if state[1] < stop[1]
-        return (stop[1],tail(state)...)
-    end
-    newtail = inc(tail(state), tail(start), tail(stop))
-    (start[1], newtail...)
-end
-@inline done{I<:CartesianIndex}(iter::CornerIterator{I}, state) = state.I[end] > iter.stop.I[end]
-
-# 0-d is special-cased to iterate once and only once
-start{I<:CartesianIndex{0}}(iter::CornerIterator{I}) = false
-next{I<:CartesianIndex{0}}(iter::CornerIterator{I}, state) = iter.start, true
-done{I<:CartesianIndex{0}}(iter::CornerIterator{I}, state) = state
+include("warp.jl")
 
 center{T,N}(img::AbstractArray{T,N}) = SVector{N}(map(_center, indices(img)))
 _center(ind::AbstractUnitRange) = (first(ind)+last(ind))/2
--- a/src/autorange.jl
+++ b/src/autorange.jl
@@ -1,0 +1,59 @@
+function autorange(img, tform)
+    R = CartesianRange(indices(img))
+    I = first(R)
+    mn = mx = x = tform(SVector(I))
+    for I in CornerIterator(R)
+        x = tform(SVector(I))
+        # we map min and max to prevent type-inference issues
+        # (because min(::SVector,::SVector) -> Vector)
+        mn = map(min, x, mn)
+        mx = map(max, x, mx)
+    end
+    _autorange(convert(Tuple, mn), convert(Tuple, mx))
+end
+
+@noinline _autorange(mn,mx) = map((a,b)->floor(Int,a):ceil(Int,b), mn, mx)
+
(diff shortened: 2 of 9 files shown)


src/autorange.jl
```diff
@@ -0,0 +1,54 @@
+function autorange(img, tform)
+    R = CartesianRange(indices(img))
+    I = first(R)
+    mn = mx = x = tform(SVector(I))
+    for I in CornerIterator(R)
+        x = tform(SVector(I))
+        # we map min and max to prevent type-inference issues
+        # (because min(::SVector,::SVector) -> Vector)
+        mn = map(min, x, mn)
+        mx = map(max, x, mx)
```
Evizero: I changed this code because of type inference issues. before it was `mn = min(x,mn)` etc

src/autorange.jl
```diff
@@ -0,0 +1,54 @@
...
+immutable CornerIterator{I<:CartesianIndex}
+    start::I
+    stop::I
+end
+CornerIterator{I<:CartesianIndex}(R::CartesianRange{I}) = CornerIterator{I}(R.start, R.stop)
+
+eltype{I}(::Type{CornerIterator{I}}) = I
+iteratorsize{I}(::Type{CornerIterator{I}}) = Base.HasShape()
+
+size{N}(iter::CornerIterator{CartesianIndex{N}}) = ntuple(d->iter.stop.I[d]-iter.start.I[d]==0 ? 1 : 2, Val{N})
```
Evizero: I added this function to complete the required iterator interface. I have yet to write formal test, but this code seems to do it and is type stable

src/warp.jl
```diff
@@ -0,0 +1,38 @@
+# None of these types are provided by this package, so the following line seems unclean
+#     @inline Base.getindex(A::AbstractExtrapolation, v::StaticVector) = A[convert(Tuple, v)...]
+# furthermore it would be ambiguous with getindex(::Extrapolation, xs...) after https://github.com/tlycken/Interpolations.jl/pull/141
+@inline _getindex(A, v::StaticVector) = A[convert(Tuple, v)...]
```
Evizero: I changed this here as well. Now that `Extrapolation <: AbstractExtrapolation` (ref: https://github.com/tlycken/Interpolations.jl/pull/141) the old code became ambiguous.

src/warp.jl
```diff
@@ -0,0 +1,38 @@
...
+
+warp(img::AbstractArray, args...) = warp(interpolate(img, BSpline(Linear()), OnGrid()), args...)
+
+function warp{T<:Colorant,S}(::Type{T}, img::AbstractArray{S}, args...)
+    TCol = typeof(convert(T,one(S)))
+    TNorm = eltype(TCol)
+    apad, pad = Interpolations.prefilter(TNorm, TCol, img, typeof(BSpline(Linear())), typeof(OnGrid()))
+    itp = Interpolations.BSplineInterpolation(TNorm, apad, BSpline(Linear()), OnGrid(), pad)
+    warp(itp, args...)
+end
```
Evizero: The way this function works may or may not be what we want. I am unsure. It is quite flexible though.

Some examples:

```julia
warp(RGB, ::Array{<:Colorant{N0f8}}) --> Array{RGB{N0f8}} 
warp(RGB{Float32}, ::Array{<:Colorant{N0f8}}) --> Array{RGB{Float32}} 
warp(Gray, ::Array{<:Colorant{N0f8}}) --> Array{Gray{N0f8}} 
warp(Gray, ::Array{<:Colorant{Float64}}) --> Array{Gray{Float64}} 
warp(Gray{N0f8}, ::Array{<:Colorant{Float64}}) --> Array{Gray{N0f8}} 
```

src/warp.jl
```diff
@@ -0,0 +1,38 @@
...
+
+@compat const FloatLike{T<:AbstractFloat} = Union{T,Gray{T}} # Why not AbstractGray ?
+@compat const FloatColorant{T<:AbstractFloat} = Colorant{T}
+
+# The default values used by extrapolation for off-domain points
+@inline _default_fill{T<:FloatLike}(::Type{T}) = convert(T, NaN)
+@inline _default_fill{T<:FloatColorant}(::Type{T}) = nan(T)
+@inline _default_fill{T}(::Type{T}) = zero(T)
+
+warp{T}(img::AbstractInterpolation{T}, tform, fill=_default_fill(T)) = warp(extrapolate(img, fill), tform)
```
Evizero: I added a third parameter to pass the extrapolation scheme / fill value. This allows things like

```julia
warp(img, tfm, Interpolations.Flat())
warp(img, tfm, 1)
```

test/autorange.jl
```diff
@@ -0,0 +1,114 @@
...
+    tst_img = zeros(h,w)
+    # We compare the result of autorange against manually computed
+    # values using basic trigonometry
+    for ϕ in (pi/8, pi/4, pi/2)
+        rot = LinearMap(RotMatrix(ϕ))
+        rnge = @inferred ImageTransformations.autorange(tst_img, rot)
+        @test rnge[1].start == -floor(sin(ϕ)*w)
+        @test rnge[1].stop  ==  floor(cos(ϕ)*h)
+        @test rnge[2].start == 1
+        @test rnge[2].stop  == ceil(cos(atan(h/w)-ϕ)*d)
```
Evizero: Can someone maybe double check these four tests? They work, but I am a little confused as of why, since the code of `autorange` applies `floor` and `ceil` differently.

Evizero:
I am playing around with updating restrict. It seems to work, but as of now my implementation is a factor 10 slower. :(

old

```
julia> @benchmark restrict($img)
BenchmarkTools.Trial: 
  memory estimate:  772.72 KiB
  allocs estimate:  32
  --------------
  minimum time:     761.340 μs (0.00% GC)
  median time:      813.791 μs (0.00% GC)
  mean time:        823.768 μs (1.89% GC)
  maximum time:     1.334 ms (34.95% GC)
  --------------
  samples:          6023
  evals/sample:     1
  time tolerance:   5.00%
  memory tolerance: 1.00%
```

new

```
julia> @benchmark restrict($img)
BenchmarkTools.Trial: 
  memory estimate:  772.97 KiB
  allocs estimate:  48
  --------------
  minimum time:     8.575 ms (0.00% GC)
  median time:      8.839 ms (0.00% GC)
  mean time:        8.868 ms (0.17% GC)
  maximum time:     9.572 ms (4.68% GC)
  --------------
  samples:          564
  evals/sample:     1
  time tolerance:   5.00%
  memory tolerance: 1.00%
```

src/warp.jl
```diff
@@ -0,0 +1,38 @@
+# None of these types are provided by this package, so the following line seems unclean
+#     @inline Base.getindex(A::AbstractExtrapolation, v::StaticVector) = A[convert(Tuple, v)...]
+# furthermore it would be ambiguous with getindex(::Extrapolation, xs...) after https://github.com/tlycken/Interpolations.jl/pull/141
+@inline _getindex(A, v::StaticVector) = A[convert(Tuple, v)...]
+
+warp(img::AbstractArray, args...) = warp(interpolate(img, BSpline(Linear()), OnGrid()), args...)
+
+function warp{T<:Colorant,S}(::Type{T}, img::AbstractArray{S}, args...)
+    TCol = typeof(convert(T,one(S)))
```
timholy: Won't `TCol == T`?

src/warp.jl
```diff
@@ -0,0 +1,38 @@
...
+    TNorm = eltype(TCol)
+    apad, pad = Interpolations.prefilter(TNorm, TCol, img, typeof(BSpline(Linear())), typeof(OnGrid()))
+    itp = Interpolations.BSplineInterpolation(TNorm, apad, BSpline(Linear()), OnGrid(), pad)
```
timholy: What are the advantages of doing this here, rather than picking types for the storage output? That is, suppose it transiently computes an `RGB{Float64}`, that can still be stashed in an `RGB{N0f8}`.

src/autorange.jl
```diff
@@ -0,0 +1,59 @@
...
+    start::I
+    stop::I
+end
+CornerIterator{I<:CartesianIndex}(R::CartesianRange{I}) = CornerIterator{I}(R.start, R.stop)
+
+eltype{I}(::Type{CornerIterator{I}}) = I
+iteratorsize{I}(::Type{CornerIterator{I}}) = Base.HasShape()
+
+# in 0.6 we could write: 1 .+ (iter.stop.I .- iter.start.I .!= 0)
+size{N}(iter::CornerIterator{CartesianIndex{N}}) = ntuple(d->iter.stop.I[d]-iter.start.I[d]==0 ? 1 : 2, Val{N})::NTuple{N,Int}
```
timholy: Hah, looks like I set to `HasSize` and then forgot to define these! :heart:

src/warp.jl
timholy: If you don't know about it already, see `ColorTypes.ccolor`.

test/autorange.jl
```diff
@@ -0,0 +1,114 @@
...
+            @test ci.start === cr.start
+            @test ci.stop === cr.stop
+            @test @inferred(eltype(ci)) === CartesianIndex{N}
+            @test @inferred(size(ci)) === ntuple(x->2, N)
+            @test @inferred(length(ci)) === 2^N
+            # Check for correct shape and values
+            A = @inferred(collect(ci))
+            @test typeof(A) <: Array{CartesianIndex{N},N}
+            if N == 1
+                @test A == [CartesianIndex{1}((si(t[1]),)), CartesianIndex{1}((ei(t[1]),))]
```
timholy: Can be written more simply as `CartesianIndex(si(t[1]))`. Likewise below for `N > 1`.

test/autorange.jl
```diff
@@ -0,0 +1,114 @@
...
+            cr = CartesianRange(t)
+            ci = @inferred(ImageTransformations.CornerIterator(cr))
+            @test typeof(ci) <: ImageTransformations.CornerIterator{CartesianIndex{N}}
+            @test ci.start === cr.start
+            @test ci.stop === cr.stop
+            @test @inferred(eltype(ci)) === CartesianIndex{N}
+            @test @inferred(size(ci)) === ntuple(x->2, N)
+            @test @inferred(length(ci)) === 2^N
+            # Check for correct shape and values
+            A = @inferred(collect(ci))
```
timholy: Lovely that you've added direct tests for these!

test/autorange.jl
timholy: Perhaps you're forgetting that the rectangle gets indexed starting at 1, not zero? So for example, the RHS of the first test should really be `floor(cos(ϕ)*1 - sin(ϕ)*w)`, and so forth. 

(Thanks for pointing this out, I appreciate your calling attention to things that pass for puzzling reasons.)

src/autorange.jl
```diff
@@ -0,0 +1,59 @@
+function autorange(img, tform)
+    R = CartesianRange(indices(img))
+    I = first(R)
+    mn = mx = x = tform(SVector(I))
```
Evizero: unsure. I just assumed there was a reason for it (since it was done explicitly before my change). I just collapsed two lines into one

src/autorange.jl
```diff
@@ -0,0 +1,59 @@
...
+immutable CornerIterator{I<:CartesianIndex}
+    start::I
+    stop::I
+end
+CornerIterator{I<:CartesianIndex}(R::CartesianRange{I}) = CornerIterator{I}(R.start, R.stop)
+
+eltype{I}(::Type{CornerIterator{I}}) = I
+iteratorsize{I}(::Type{CornerIterator{I}}) = Base.HasShape()
+
+# in 0.6 we could write: 1 .+ (iter.stop.I .- iter.start.I .!= 0)
```
Evizero: Good to know. Will do

src/warp.jl
Evizero: Not quite. I do this only for user convenience, so they don't need to specify fully qualified Colorants

```julia
julia> T = Gray
ColorTypes.Gray{T<:Union{AbstractFloat,Bool,FixedPointNumbers.FixedPoint}}

julia> S = RGB{N0f8}
ColorTypes.RGB{FixedPointNumbers.Normed{UInt8,8}}

julia> TCol = typeof(convert(T,one(S)))
ColorTypes.Gray{FixedPointNumbers.Normed{UInt8,8}}
```

src/warp.jl
Evizero: Do you mean leaving the eltype of the interpolation alone, but instead overwrite the eltype of the `out` array? I thought about this as well, I just hoped that maybe by pushing the convert responsibility to `Iterpolations`, that it can optimize for it.

Should we change?

src/warp.jl
Evizero: did not know about `ccolor`. Will adapt

src/warp.jl
timholy: As long as you're pretty confident in the code, this can stay. I was mostly curious, since it's more work this way (but potentially faster I guess).

src/ImageTransformations.jl
```diff
@@ -1,93 +1,28 @@
...
 
-    warp,
-    center,
     restrict,
-    imresize
+    imresize,
+    warp,
+    center
 
+include("autorange.jl")
```
Evizero: I had the same internal struggle. I even named it `corneriterator.jl` at first.  I then went with `autorange.jl` since it is the only function using `CornerIterator`, while other functions use `autorange`. A bit arbitrary, but at least it is some kind of reason :)

Evizero:
I think I addressed everything. ready for another round of review and/or merge

test/resizing.jl
```diff
@@ -0,0 +1,77 @@
+@testset "Restriction" begin
+    imgcol = colorview(RGB, rand(3,5,6))
```
timholy: Not used?

Evizero: oh yea, seems to be a relic from the axis array tests, which moved to Images.jl

timholy:
Thanks. I don't need to see this get through CI again, so I say merge. I'll let you decide whether this should be squashed in one big commit or whether smaller commits are more appropriate.

Evizero:
I think a squash merge is the only sensible way to go for changes like this one, since it is such diff-breaking code move

timholy:
Thanks for doing this! :fireworks:

Evizero:
happy to help!

src/autorange.jl
```diff
@@ -0,0 +1,58 @@
...
+    stop::I
+end
+CornerIterator{I<:CartesianIndex}(R::CartesianRange{I}) = CornerIterator{I}(first(R), last(R))
+
+eltype{I}(::Type{CornerIterator{I}}) = I
+iteratorsize{I}(::Type{CornerIterator{I}}) = Base.HasShape()
+
+# in 0.6 we could write: 1 .+ (iter.stop.I .- iter.start.I .!= 0)
+size{N}(iter::CornerIterator{CartesianIndex{N}}) = ntuple(d->iter.stop.I[d]-iter.start.I[d]==0 ? 1 : 2, Val{N})::NTuple{N,Int}
+length(iter) = prod(size(iter))
```
tkelman: Should this have a type constraint on it?

Evizero: oh boy. good catch.

PR #12, opened 2017-03-16
prototype new restrict based on CartesianIndex

Evizero:
First of all, sorry for the constant noise! I felt it more appropriate to create a PR for just the restrict refactor.

Basically I reimplemented the `restrict!` function using `CartesianIndex` instead of the `Base.Cartesian` macros. I renamed the old implementation to `restrict_old!`.

I managed to get the performance within a factor two/three of the original implementation. I also added regression tests that compare the result against the old implementation

Here are some benchmarks

### Lena image

```julia
julia> @benchmark ImageTransformations.restrict_old($(testimage("lena")))
BenchmarkTools.Trial: 
  memory estimate:  582.84 KiB
  allocs estimate:  40
  --------------
  minimum time:     473.245 μs (0.00% GC)
  median time:      490.146 μs (0.00% GC)
  mean time:        505.872 μs (2.49% GC)
  maximum time:     1.137 ms (47.31% GC)
  --------------
  samples:          9769
  evals/sample:     1
  time tolerance:   5.00%
  memory tolerance: 1.00%

julia> @benchmark ImageTransformations.restrict($(testimage("lena")))
BenchmarkTools.Trial: 
  memory estimate:  582.47 KiB
  allocs estimate:  16
  --------------
  minimum time:     982.108 μs (0.00% GC)
  median time:      1.025 ms (0.00% GC)
  mean time:        1.035 ms (1.16% GC)
  maximum time:     1.620 ms (30.35% GC)
  --------------
  samples:          4800
  evals/sample:     1
  time tolerance:   5.00%
  memory tolerance: 1.00%
```

### Camerman

```julia
julia> @benchmark ImageTransformations.restrict_old($(testimage("camera")))
BenchmarkTools.Trial: 
  memory estimate:  772.72 KiB
  allocs estimate:  32
  --------------
  minimum time:     767.346 μs (0.00% GC)
  median time:      790.394 μs (0.00% GC)
  mean time:        813.456 μs (2.02% GC)
  maximum time:     1.430 ms (34.61% GC)
  --------------
  samples:          6100
  evals/sample:     1
  time tolerance:   5.00%
  memory tolerance: 1.00%

julia> @benchmark ImageTransformations.restrict($(testimage("camera")))
BenchmarkTools.Trial: 
  memory estimate:  772.72 KiB
  allocs estimate:  32
  --------------
  minimum time:     2.495 ms (0.00% GC)
  median time:      2.574 ms (0.00% GC)
  mean time:        2.595 ms (0.64% GC)
  maximum time:     3.213 ms (15.06% GC)
  --------------
  samples:          1922
  evals/sample:     1
  time tolerance:   5.00%
  memory tolerance: 1.00%
```

### DIFF
--- a/src/resizing.jl
+++ b/src/resizing.jl
@@ -23,111 +23,67 @@ function restrict{T,N}(A::AbstractArray{T,N}, dim::Integer)
     end
     newsz = ntuple(i->i==dim?restrict_size(size(A,dim)):size(A,i), Val{N})
     # FIXME: The following line fails for interpolations because
-    # interpolations can be accessed linearily A[i].
+    # interpolations can not be accessed linearily A[i].
     #    out = Array{typeof(A[1]/4+A[2]/2),N}(newsz)
     out = Array{typeof(first(A)/4+first(A)/2),N}(newsz)
     restrict!(out, A, dim)
     out
 end
 
-# out should have efficient linear indexing
-for N = 1:5
-    @eval begin
-        function restrict!{T}(out::AbstractArray{T,$N}, A::AbstractArray, dim)
-            if isodd(size(A, dim))
-                half = convert(eltype(T), 0.5)
-                quarter = convert(eltype(T), 0.25)
-                indx = 0
-                if dim == 1
-                    @inbounds @nloops $N i d->(d==1 ? (1:1) : (1:size(A,d))) d->(j_d = d==1 ? i_d+1 : i_d) begin
-                        nxt = convert(T, @nref $N A j)
-                        out[indx+=1] = half*(@nref $N A i) + quarter*nxt
-                        for k = 3:2:size(A,1)-2
-                            prv = nxt
-                            i_1 = k
-                            j_1 = k+1
-                            nxt = convert(T, @nref $N A j)
-                            out[indx+=1] = quarter*(prv+nxt) + half*(@nref $N A i)
-                        end
-                        i_1 = size(A,1)
-                        out[indx+=1] = quarter*nxt + half*(@nref $N A i)
-                    end
-                else
-                    strd = stride(out, dim)
-                    # Must initialize the i_dim==1 entries with zero
-                    @nexprs $N d->sz_d=d==dim?1:size(out,d)
-                    @nloops $N i d->(1:sz_d) begin
-                        (@nref $N out i) = zero(T)
-                    end
-                    stride_1 = 1
-                    @nexprs $N d->(stride_{d+1} = stride_d*size(out,d))
-                    @nexprs $N d->offset_d = 0
-                    ispeak = true
-                    @inbounds @nloops $N i d->(d==1?(1:1):(1:size(A,d))) d->(if d==dim; ispeak=isodd(i_d); offset_{d-1} = offset_d+(div(i_d+1,2)-1)*stride_d; else; offset_{d-1} = offset_d+(i_d-1)*stride_d; end) begin
-                        indx = offset_0
-                        if ispeak
-                            for k = 1:size(A,1)
-                                i_1 = k
-                                out[indx+=1] += half*(@nref $N A i)
-                            end
-                        else
-                            for k = 1:size(A,1)
-                                i_1 = k
-                                tmp = quarter*(@nref $N A i)
-                                out[indx+=1] += tmp
-                                out[indx+strd] = tmp
-                            end
-                        end
-                    end
-                end
+function restrict!{T,S,N}(out::AbstractArray{T,N}, A::AbstractArray{S}, dim)
+    # one_d is a tuple in which all elements are 0 expect for "dim"
+    #   e.g. for N=4 and dim=3: one_d=CartesianIndex((0,0,1,0))
+    one_d = CartesianIndex{N}(ntuple(d->d==dim?1:0, Val{N})::NTuple{N,Int})
+    # mlp is a tuple of multiplier in which all elements are 1 expect for "dim"
+    #   e.g. for N=4 and dim=3: mlp=(1,1,2,1)
+    mlp = ntuple(d->d==dim?2:1, Val{N})::NTuple{N,Int}
+    inds = indices(out)
+    min_d, max_d = extrema(inds[dim])
+    if isodd(length(indices(A,dim)))
+        half    = convert(eltype(T), 0.5)
+        quarter = convert(eltype(T), 0.25)
+        @inbounds for O in CartesianRange(inds)
+            # compute corresponding index in A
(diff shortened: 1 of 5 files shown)


src/restrict_old.jl
```diff
@@ -0,0 +1,121 @@
...
+    # FIXME: The following line fails for interpolations because
+    # interpolations can not be accessed linearily A[i].
+    #    out = Array{typeof(A[1]/4+A[2]/2),N}(newsz)
+    out = Array{typeof(first(A)/4+first(A)/2),N}(newsz)
+    restrict_old!(out, A, dim)
+    out
+end
+
+# out should have efficient linear indexing
+for N = 1:5
```
timholy: We could probably just turn this into a `@generated` function.

src/resizing.jl
```diff
@@ -23,111 +23,67 @@ function restrict{T,N}(A::AbstractArray{T,N}, dim::Integer)
     end
     newsz = ntuple(i->i==dim?restrict_size(size(A,dim)):size(A,i), Val{N})
     # FIXME: The following line fails for interpolations because
-    # interpolations can be accessed linearily A[i].
+    # interpolations can not be accessed linearily A[i].
```
timholy: a rather important clarification!

src/resizing.jl
```diff
@@ -23,111 +23,67 @@ function restrict{T,N}(A::AbstractArray{T,N}, dim::Integer)
...
-                                out[indx+=1] += tmp
-                                out[indx+strd] = tmp
-                            end
-                        end
-                    end
-                end
+function restrict!{T,S,N}(out::AbstractArray{T,N}, A::AbstractArray{S}, dim)
+    # one_d is a tuple in which all elements are 0 expect for "dim"
+    #   e.g. for N=4 and dim=3: one_d=CartesianIndex((0,0,1,0))
+    one_d = CartesianIndex{N}(ntuple(d->d==dim?1:0, Val{N})::NTuple{N,Int})
```
timholy: If this isn't inferred, you might be running up against https://github.com/JuliaLang/julia/issues/15276? Do `dim`-dependent operations exhibit type-instability anywhere in here?

src/resizing.jl
```diff
@@ -23,111 +23,67 @@ function restrict{T,N}(A::AbstractArray{T,N}, dim::Integer)
...
+    mlp = ntuple(d->d==dim?2:1, Val{N})::NTuple{N,Int}
+    inds = indices(out)
+    min_d, max_d = extrema(inds[dim])
+    if isodd(length(indices(A,dim)))
+        half    = convert(eltype(T), 0.5)
+        quarter = convert(eltype(T), 0.25)
+        @inbounds for O in CartesianRange(inds)
+            # compute corresponding index in A
+            I = CartesianIndex{N}(map(*, O.I, mlp)) - one_d
+            if O[dim] == min_d
```
timholy: Part of the performance of the previous version was avoiding branches in the inner loop. If you want to pursue this, one option would be to split the first dimension out from the rest.

timholy:
> First of all, sorry for the constant noise!

Don't be! Sorry I'm slow to give feedback, I'm up to my neck in several other changes ranging from indexing rules in Base to graphics in GtkReactive.

It's a very impressive PR, thanks for doing this.

src/resizing.jl
Evizero: Unsure. Lately I regularly run into weird inference issues in all sorts of packages I am working on, where everything with `@inferred` works fine in the REPL, but `Pkg.test` fails. I know that there are some additional flags set in `Pkg.test`, but I have not yet been frustrated enough to investigate

Evizero: Oh, and yea, I think the inference failed just for some value of `dim` if I remember correctly. Maybe I should look into this a little closer...

Evizero:
I made an interesting observation

Something is going on with the old implementation in the `isodd` branch, which seems vastly inferior to the `else` branch. The new implementation at least beats that time :). Unless I am doing something stupid here, which is always a lingering possibility 

### odd size

```julia
julia> img = testimage("cameraman")[1:511,1:511];

julia> out = zeros(restrict(img, 1));

julia> @benchmark ImageTransformations.restrict!($out, $img, 1)
BenchmarkTools.Trial: 
  memory estimate:  0 bytes
  allocs estimate:  0
  --------------
  minimum time:     1.432 ms (0.00% GC)
  median time:      1.488 ms (0.00% GC)
  mean time:        1.491 ms (0.00% GC)
  maximum time:     1.794 ms (0.00% GC)
  --------------
  samples:          3342
  evals/sample:     1
  time tolerance:   5.00%
  memory tolerance: 1.00%

julia> @benchmark ImageTransformations.restrict_old!($out, $img, 1)
BenchmarkTools.Trial: 
  memory estimate:  15.97 MiB
  allocs estimate:  1046528
  --------------
  minimum time:     16.540 ms (0.00% GC)
  median time:      17.937 ms (6.53% GC)
  mean time:        17.794 ms (4.87% GC)
  maximum time:     19.922 ms (0.00% GC)
  --------------
  samples:          281
  evals/sample:     1
  time tolerance:   5.00%
  memory tolerance: 1.00%
```

### even size

```julia
julia> img = testimage("cameraman");

julia> out = zeros(restrict(img, 1));

julia> @benchmark ImageTransformations.restrict!($out, $img, 1)
BenchmarkTools.Trial: 
  memory estimate:  0 bytes
  allocs estimate:  0
  --------------
  minimum time:     2.019 ms (0.00% GC)
  median time:      2.102 ms (0.00% GC)
  mean time:        2.096 ms (0.00% GC)
  maximum time:     2.457 ms (0.00% GC)
  --------------
  samples:          2379
  evals/sample:     1
  time tolerance:   5.00%
  memory tolerance: 1.00%

julia> @benchmark ImageTransformations.restrict_old!($out, $img, 1)
BenchmarkTools.Trial: 
  memory estimate:  0 bytes
  allocs estimate:  0
  --------------
  minimum time:     697.086 μs (0.00% GC)
  median time:      726.350 μs (0.00% GC)
  mean time:        723.565 μs (0.00% GC)
  maximum time:     893.759 μs (0.00% GC)
  --------------
  samples:          6858
  evals/sample:     1
  time tolerance:   5.00%
  memory tolerance: 1.00%
```

Evizero:
I doubt I'll get any closer that 2x to the original implementation. I tried simulating the branch using some tricks, but that made it worse actually. I agree that performance trumps readability.

I'll close and abandon this PR for now, but it was a fun exercise. I'll leave the branch for future reference if we ever want to revisit the idea (I'll rename it to something like `restrict_CartesianIndex`).

PR #13, opened 2017-03-16
RFC: WarpedView: lazy version of warp. Alternative to TransformedArray

Evizero:
I am experimenting with the idea of a view version of `warp`. this is a raw draft, but it works.
I am unsure if this package is the right place, but it can't hurt to throw the idea out there

- Ideally I would like to allow it to unroll nested `WarpedView` calls by composing transformations. 

- Maybe `warpedview` could dispatch on `::WarpedView` and compose the transformations instead of nesting views. Seems to be in line with the `ImageCore` design.

any thoughts on this?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,9 +9,9 @@ julia:
 notifications:
   email: false
 # uncomment the following lines to override the default test script
-#script:
-#  - if [[ -a .git/shallow ]]; then git fetch --unshallow; fi
-#  - julia -e 'Pkg.clone(pwd()); Pkg.build("ImageTransformations"); Pkg.test("ImageTransformations"; coverage=true)'
+script:
+  - if [[ -a .git/shallow ]]; then git fetch --unshallow; fi
+  - julia -e 'Pkg.clone(pwd()); Pkg.build("ImageTransformations"); Pkg.test("ImageTransformations"; coverage=VERSION >= v"0.6.0-alpha")'
 after_success:
   # push coverage results to Codecov
-  - julia -e 'cd(Pkg.dir("ImageTransformations")); Pkg.add("Coverage"); using Coverage; Codecov.submit(Codecov.process_folder())'
+  - julia -e 'if VERSION >= v"0.6.0-alpha" cd(Pkg.dir("ImageTransformations")); Pkg.add("Coverage"); using Coverage; Codecov.submit(Codecov.process_folder()); end'
--- a/src/warpedview.jl
+++ b/src/warpedview.jl
@@ -11,8 +11,17 @@ immutable WarpedView{T,N,A<:AbstractArray,F1<:Transformation,F2<:Transformation,
     end
 end
 
-function WarpedView(parent::AbstractArray, args...)
-    WarpedView(interpolate(parent, BSpline(Linear()), OnGrid()), args...)
+function WarpedView{T,N,F<:Transformation}(inner::WarpedView{T,N}, tform_in::F)
+    tform = compose(tform_in, inner.transform)
+    tinv = inv(tform)
+    etp = inner.parent
+    inds = autorange(etp, tform)
+    WarpedView{T,N,typeof(etp),typeof(tform),typeof(tinv),typeof(inds)}(etp, tform, tinv, inds)
+end
+
+function WarpedView{T,N}(parent::AbstractArray{T,N}, args...)
+    itp = Interpolations.BSplineInterpolation{T,N,typeof(parent),Interpolations.BSpline{Interpolations.Linear},OnGrid,0}(parent)
+    WarpedView(itp, args...)
 end
 
 function WarpedView{T,F<:Transformation}(itp::AbstractInterpolation{T}, tform::F, fill=_default_fill(T))
@@ -25,11 +34,26 @@ function WarpedView{T,N,F<:Transformation}(etp::AbstractExtrapolation{T,N}, tfor
     WarpedView{T,N,typeof(etp),F,typeof(tinv),typeof(inds)}(etp, tform, tinv, inds)
 end
 
-Base.parent(A::WarpedView)  = A.parent
+Base.parent(A::WarpedView) = A.parent
 @inline Base.indices(A::WarpedView) = A.indices
 
 @compat Compat.IndexStyle{T<:WarpedView}(::Type{T}) = IndexCartesian()
-@inline Base.getindex{N}(A::WarpedView, I::Vararg{Int,N}) = _getindex(A.parent, A.transform_inv(SVector(I)))
+@inline Base.getindex{T,N}(A::WarpedView{T,N}, I::Vararg{Int,N}) =
+    _getindex(A.parent, A.transform_inv(SVector(I)))
 
 Base.size(A::WarpedView)    = OffsetArrays.errmsg(A)
 Base.size(A::WarpedView, d) = OffsetArrays.errmsg(A)
+
+function ShowItLikeYouBuildIt.showarg(io::IO, A::WarpedView)
+    print(io, "WarpedView(")
+    showarg(io, _parent(A))
+    print(io, ", ")
+    showarg(io, A.transform)
+    print(io, ')')
+end
+
+_parent(A) = parent(A)
+_parent(A::WarpedView) = _parent(parent(A))
+_parent(A::AbstractInterpolation) = _parent(parent(A))
+
+Base.summary(A::WarpedView) = summary_build(A)
--- a/src/warpedview.jl
+++ b/src/warpedview.jl
@@ -46,14 +46,10 @@ Base.size(A::WarpedView, d) = OffsetArrays.errmsg(A)
 
 function ShowItLikeYouBuildIt.showarg(io::IO, A::WarpedView)
     print(io, "WarpedView(")
-    showarg(io, _parent(A))
+    showarg(io, parent(A))
     print(io, ", ")
-    showarg(io, A.transform)
+    print(io, A.transform)
     print(io, ')')
 end
-
-_parent(A) = parent(A)
-_parent(A::WarpedView) = _parent(parent(A))
-_parent(A::AbstractInterpolation) = _parent(parent(A))
 
 Base.summary(A::WarpedView) = summary_build(A)
--- a/src/warpedview.jl
+++ b/src/warpedview.jl
@@ -1,0 +1,35 @@
+immutable WarpedView{T,N,A<:AbstractArray,F1<:Transformation,F2<:Transformation,I} <: AbstractArray{T,N}
+    parent::A
+    transform::F1
+    transform_inv::F2
+    indices::I
+
+    function (::Type{WarpedView{T,N,A,F1,F2,I}}){T,N,A<:AbstractArray,F1<:Transformation,F2<:Transformation,I}(
+            parent::A, tform::F1, tinv::F2, indices::I)
+        @assert eltype(parent) == T
+        new{T,N,A,F1,F2,I}(parent, tform, tinv, indices)
+    end
+end
+
+function WarpedView(parent::AbstractArray, args...)
+    WarpedView(interpolate(parent, BSpline(Linear()), OnGrid()), args...)
+end
+
+function WarpedView{T,F<:Transformation}(itp::AbstractInterpolation{T}, tform::F, fill=_default_fill(T))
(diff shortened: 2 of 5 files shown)


timholy:
Yeah, I like it. There's some undocumented functionality doing this in AffineTransforms.jl. Dunno if it's helpful, could even be confusing. In particular, IIRC for linear interpolation there's some impressive-if-I-say-so-myself optimizations (to avoid calculating `NaN` pixels), but that kind of trickery is an absolutely wonderful source of bugs, and decreases maintainability and flexibility.

timholy:
Oh, and I do think this package is the right place for such things. Really nice!

Evizero:
> There's some undocumented functionality doing this in AffineTransforms.jl.

Yes. As far as I remember the idea was pitched to rewrite `TransformedArray` using `CoordinateTransformations` and friends.

> IIRC for linear interpolation there's some impressive-if-I-say-so-myself optimizations (to avoid calculating NaN pixels),

I'll look into it. I thought NaN pixels through of domain extrapolations are considered a feature? or is there a different source of NaNs that I am ignorant of?

timholy:
> As far as I remember the idea was pitched to rewrite TransformedArray using CoordinateTransformations and friends.

Absolutely.

> I thought NaN pixels caused by off-domain extrapolations are considered a feature? or is there a different source of NaNs that I am ignorant of?

Yes, but if you know the result is going to be `NaN` based on bounds, you can save some computation if you're iterating with a deterministic pattern. IIRC, when iterating over dimension 1 it first computes the `start:stop` range along that dimension that corresponded to the in-bounds region of the original array; conceptually, rather than
```julia
    for i = 1:n
        if !in_bounds(inds)
            out[blah...] = NaN
        else
            out[blah...] = compute_something(img, blah...)
        end
    end
```
it's
```julia
    for i = frst:lst
        # frst and lst have been chosen to be in_bounds
        out[blah...] = compute_something(img, blah...)
    end
```
This saves a branch in the inner loop (a common theme, you can see).

I'm not advocating you go there, however, just explaining why the code in AffineTransforms is complicated.

Evizero:
Oh that is interesting. Actually wrapping my head around domain limits is quite important to me, because I have an actual use case where I would like to keep track of the in-domain pixel of the image after a bunch of transformations. 

For example I want to be able to rotate the image and then take a random crop of maximal size that has a defined aspect ration and only contains in-domain pixel.

I am not quite there yet, but your particular suggestion seems like a good breadcrumb to get the ball rolling

Evizero:
Any suggestion/opinion on the name `WarpedView` in general? I am trying to be consistent with the function name `warp`. Where does the function name `warp` originate?

timholy:
`warp` comes from the notion that a general coordinate transformation `img[x] -> img[ϕ(x)]` will cause "warping" of the image. (E.g., https://en.wikipedia.org/wiki/Image_warping.) "Transformation" is pretty general: changing the colors would be a form of transformation that doesn't affect space in any way, so "warp" is a good term if you want to be specific about spatial transformations.

:+1: for `WarpedView`.

Evizero:
I have to admit I am getting rather annoyed by this:

screen shot 2017-04-15 at 17 38 08

I am looking into `ShowItLikeYouBuildIt` to find a better way to display it.

Are you open to the idea of "hiding" the extrapolation/intrapolation from the type?

Evizero:
This could be one way to approach it. Though I have to admit it might hide a bit too much information.

screen shot 2017-04-15 at 19 15 00

src/warpedview.jl
```diff
@@ -0,0 +1,59 @@
...
+function WarpedView{T,N,F<:Transformation}(inner::WarpedView{T,N}, tform_in::F)
+    tform = compose(tform_in, inner.transform)
+    tinv = inv(tform)
+    etp = inner.parent
+    inds = autorange(etp, tform)
+    WarpedView{T,N,typeof(etp),typeof(tform),typeof(tinv),typeof(inds)}(etp, tform, tinv, inds)
+end
+
+function WarpedView{T,N}(parent::AbstractArray{T,N}, args...)
+    itp = Interpolations.BSplineInterpolation{T,N,typeof(parent),Interpolations.BSpline{Interpolations.Linear},OnGrid,0}(parent)
```
Evizero: I create the interpolation this way, because of two orthogonal reasons. I'd be curious about your opinion on this.

1. `interpolate` calls `prefilter` (see: https://github.com/JuliaMath/Interpolations.jl/blob/master/src/b-splines/b-splines.jl#L63-L66) which causes the resulting `BSplineInterpolation` to be a decorator around a **copy** of `parent`. This goes a bit against the idea of a view. On the other hand I am guessing that there is a good reason to call `prefilter`.

2. A call to `interpolate` usually causes the resulting `BSplineInterpolation` to have a different element type. Given the meaning behind `warp` (that you were nice enough to explain to me here https://github.com/JuliaImages/ImageTransformations.jl/pull/13#issuecomment-292722621), I think we should be adamant about **only** performing a warp.

Evizero: Given point 2., I am inclined to think we should make `warp` behave the same way by default. Changing the `eltype` seems like an orthogonal task to warping. Plus the user has the option to provide a custom `Interpolation` anyways

timholy: You're right, another easy way is to use `mappedarray` on the input.

timholy: The point of `prefilter` is that interpolation beyond `Linear` order requires a different coefficient matrix than the input values. For example:
```julia
julia> using Interpolations

julia> a = 1:5
1:5

julia> ai = interpolate(a, BSpline(Quadratic(Flat())), OnCell())
5-element Interpolations.BSplineInterpolation{Float64,1,Array{Float64,1},Interpolations.BSpline{Interpolations.Quadratic{Interpolations.Flat}},Interpolations.OnCell,1}:
 1.0
 2.0
 3.0
 4.0
 5.0

julia> ai.coefs
7-element Array{Float64,1}:
 0.853659
 0.853659
 2.02439 
 3.0     
 3.97561 
 5.14634 
 5.14634 
```
What `prefilter` does is essentially "invert" the computation (at the grid points) that interpolation will perform. For quadratic interpolation, the output for on-grid interpolation is given by
```julia
out[i] = (1/8)*coef[i-1] + (3/4)*coef[i] + (1/8)*coef[i+1]
```
and so the coefficients are calculated by inverting a tridiagonal system of equations.

Evizero: I see. Thank you. So if I hardcode `Linear` and `OnGrid` and set pad to `0`, is it safe to bypass `prefilter`? I am trying to preserve the nature of a view in a clean way

timholy:
I agree the wall of text is impossible to read. I'm open to the idea of hiding all details about the interpolation order, although a disadvantage is that one doesn't know the order of the interpolant (`Constant`, `Linear`, `Quadratic`, etc). What if it displayed something like this?
```
-78:591x-78:591 WarpedView(extrapolate(interpolate(::Array{Gray{N0f8},2}, BSpline(Linear()), OnGrid()), 0), <transformation>) with element type Gray{Float64}
```
I'm not quite sure how to display the transformation in the summary string; the type-based method you showed above would be fine, but for Affine transformations we could alternatively consider showing the transformation matrix itself.

Evizero:
I considered the `extrapolate(` style as well. The question that comes up here is where to define that behaviour. Should we try to propose the idea to introduce `ShowItLikeYouBuildIt` in `Interpolations` directly, or do some kind of special treatment in this package?

src/warpedview.jl
timholy: Yes, those should be fine. You're right that the name `WarpedView` really does mean that it shouldn't support interpolation schemes that require modifications to the coefficient matrix. :+1:

timholy:
Let's float the idea with the watchers of Interpolations, and if they don't like it then let's add it here. This is a case of deliberate type-piracy, essentially as an experiment in how one displays complex nested objects (since I think Images and the various AbstractArray types it uses is pushing the frontier of nested types).

Evizero:
I'll code up a PR.

src/warpedview.jl
timholy: Though of course if the user supplies an `itp`/`etp` object, then it's fine to consider it a view since it's a view of what's being passed in as an argument.

Evizero:
How about this? (depends on https://github.com/JuliaMath/Interpolations.jl/pull/148)

screen shot 2017-04-16 at 18 46 58

Evizero:
I still want to look into the test timeouts a bit and see if I can't speed up the testing time somewhat.

I should come up with a checklist or something

- [ ] Docstring for `WarpedView`
- [ ] Make tests not time out

anything else? Concerning the refactor or `warp` to the same "warp only" principle discussed in https://github.com/JuliaImages/ImageTransformations.jl/pull/13#discussion_r111682755, I'll do that in a separate PR.

timholy:
Yes, the test timeout does need to be solved. Personally, I would just turn off coverage on anything less than 0.6.

And the docstring would be great. Thanks for noticing that, I had missed it.

I guess the other thing is that the test coverage for this patch is currently claimed to be relatively low, though I'm not sure I believe that.

andyferris:
I'm not 100% sold that the entire type is necessary in the header.

Personally, I'd be happy with `-78:591×-78×591 WarpedView:` as the header. I can use `typeof` when I want that extra information.

The main advantage of the default behavior for `Array` is so that you can see the element type, so we might want to throw that in. For more complex arrays it's usually not useful information at the REPL (except as a package developer doing debugging).

Evizero:
I have to say I have grown to really like the fact that the transformation is displayed. One thing we could try is hide the whole extrapolation information if standard settings are used.

instead of

-78:591×-78:591 WarpedView(extrapolate(interpolate(::Array{Gray{N0f8},2}, BSpline(Linear()), OnGrid()), Gray{N0f8}(0.0)), AffineMap([0.92388 0.382683; -0.382683 0.92388], [-78.6334,117.683])) with element type ColorTypes.Gray{FixedPointNumbers.Normed{UInt8,8}}:

we do

-78:591×-78:591 WarpedView(::Array{Gray{N0f8},2}, AffineMap([0.92388 0.382683; -0.382683 0.92388], [-78.6334,117.683])) with element type ColorTypes.Gray{FixedPointNumbers.Normed{UInt8,8}}:

if `BSpline(Linear())` with `OnGrid()` is in use

Evizero:
Like this:

default settings when passing an array:

screen shot 2017-04-21 at 09 33 34

when passing a custom interpolation with other settings than default (but still default extrapolation, which is hidden):

screen shot 2017-04-21 at 09 34 31

custom extrapolation as well:

screen shot 2017-04-21 at 09 38 29

Evizero:
If this is ok, I'd like to merge this PR as it is (without a docstring).

I have some ideas to change the API slightly, which requires some other things that are a bit orthogonal to the main idea behind this PR, and also maybe a bit controversial. So I'd like to do it in a separate PR, where we can start fresh based on what we have now.

This also means that I'd like to delay a new Tag a bit, because I'd prefer to not commit to the current API of `WarpedView` just yet.

.travis.yml
```diff
@@ -9,9 +9,9 @@ julia:
 notifications:
   email: false
 # uncomment the following lines to override the default test script
-#script:
-#  - if [[ -a .git/shallow ]]; then git fetch --unshallow; fi
-#  - julia -e 'Pkg.clone(pwd()); Pkg.build("ImageTransformations"); Pkg.test("ImageTransformations"; coverage=true)'
+script:
+  - if [[ -a .git/shallow ]]; then git fetch --unshallow; fi
+  - julia -e 'Pkg.clone(pwd()); Pkg.build("ImageTransformations"); Pkg.test("ImageTransformations"; coverage=VERSION >= v"0.6.0-alpha")'
```
tkelman: pre.alpha

timholy: Seems to be working OK, though? https://codecov.io/gh/JuliaImages/ImageTransformations.jl
Once 0.6 is out we can drop the alpha part.

tkelman: what are you testing for? the very first dev build after 0.5 branched is > alpha

timholy: We only want travis to run coverage on julia-0.6, because when you turn off inlining it's too slow. Since this works and the only "user" affected is Travis, I suspect we're OK.

tkelman: yes it's not a big deal since it's on travis, but this is an incorrect version comparison and shouldn't be used anywhere else

PR #14, opened 2017-03-16
make restrict a generated function and fix convert bug

Evizero:
I am afraid the indentation change makes the diff very hard to read.

basically I only did two things:

1. made `restrict` a generated function
2. made sure that all `@nref $N A i` are surrounded by `convert(T,...)` this seems to fix the huge performance overhead described at https://github.com/JuliaImages/ImageTransformations.jl/pull/12#issuecomment-287180807

```julia
julia> img = testimage("cameraman")[1:511,1:511];

julia> out = zeros(restrict(img, 1));

julia> @benchmark ImageTransformations.restrict!($out, $img, 1)
BenchmarkTools.Trial: 
  memory estimate:  0 bytes
  allocs estimate:  0
  --------------
  minimum time:     865.264 μs (0.00% GC)
  median time:      909.404 μs (0.00% GC)
  mean time:        903.599 μs (0.00% GC)
  maximum time:     1.732 ms (0.00% GC)
  --------------
  samples:          5499
  evals/sample:     1
  time tolerance:   5.00%
  memory tolerance: 1.00%
```

🎉

### DIFF
--- a/src/resizing.jl
+++ b/src/resizing.jl
@@ -31,97 +31,95 @@ function restrict{T,N}(A::AbstractArray{T,N}, dim::Integer)
 end
 
 # out should have efficient linear indexing
-for N = 1:5
-    @eval begin
-        function restrict!{T}(out::AbstractArray{T,$N}, A::AbstractArray, dim)
-            if isodd(size(A, dim))
-                half = convert(eltype(T), 0.5)
-                quarter = convert(eltype(T), 0.25)
-                indx = 0
-                if dim == 1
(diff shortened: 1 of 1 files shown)


timholy:
Awesome! I poked at this a bit myself, and noticed that the problem, curiously, is `*(x::Float32, y::N0f8) = *(promote(x, y)...)`. Doing the type-conversion manually clearly solves that problem. I think we can merge this, but I will also see what I can do about fixing this independently in FixedPointNumbers.

> I am afraid the indentation change makes the diff very hard to read.

Cool github tip: add `?w=1` to the end of the URL when looking at the "Files changed" tab.

Evizero:
> Cool github tip: add ?w=1 to the end of the URL when looking at the "Files changed" tab

Nice! Definitely my tip of the day

PR #18, opened 2017-04-11
Support arbitrary indices

timholy:
This leverages recent changes in Interpolations (https://github.com/JuliaMath/Interpolations.jl/pull/146) to provide support for arbitrary indices in warping. It also rewrites `restrict` to support arbitrary indices. Note that this very similar to https://github.com/JuliaImages/ImageTransformations.jl/pull/12, but by keeping any branches out of the innermost loop and by circumventing https://github.com/JuliaLang/julia/issues/9080, I've been able to keep the performance essentially unchanged (less than 10% difference).

### DIFF
--- a/src/resizing.jl
+++ b/src/resizing.jl
@@ -10,125 +10,148 @@ See also [`imresize`](@ref).
 """
 restrict(img::AbstractArray, ::Tuple{}) = img
 
-function restrict(A::AbstractArray, region::Union{Dims,Vector{Int}} = coords_spatial(A))
-    for dim in region
-        A = restrict(A, dim)
-    end
-    A
+restrict(A::AbstractArray, region::Vector{Int}) = restrict(A, (region...))
+function restrict(A::AbstractArray, region::Dims = coords_spatial(A))
+    restrict(restrict(A, region[1]), Base.tail(region))
 end
 
 function restrict{T,N}(A::AbstractArray{T,N}, dim::Integer)
-    if size(A, dim) <= 2
-        return A
-    end
-    newsz = ntuple(i->i==dim?restrict_size(size(A,dim)):size(A,i), Val{N})
-    # FIXME: The following line fails for interpolations because
-    # interpolations can be accessed linearily A[i].
-    #    out = Array{typeof(A[1]/4+A[2]/2),N}(newsz)
-    out = Array{typeof(first(A)/4+first(A)/2),N}(newsz)
+    indsA = indices(A)
+    newinds = ntuple(i->i==dim?restrict_indices(indsA[dim]):indsA[i], Val{N})
+    out = similar(Array{typeof(first(A)/4+first(A)/2),N}, newinds)
     restrict!(out, A, dim)
     out
 end
 
-# out should have efficient linear indexing
-@generated function restrict!{T,N}(out::AbstractArray{T,N}, A::AbstractArray, dim)
+function restrict!{T,N}(out::AbstractArray{T,N}, A::AbstractArray, dim)
+    if dim > N
+        return copy!(out, A)
+    end
+    indsout, indsA = indices(out), indices(A)
+    ndims(out) == ndims(A) || throw(DimensionMismatch("input and output must have the same number of dimensions"))
+    for d = 1:length(indsA)
+        target = d==dim ? restrict_indices(indsA[d]) : indsA[d]
+        indsout[d] == target || error("input and output must have corresponding indices; to be consistent with the input indices,\ndimension $d should be $target, got $(indsout[d])")
+    end
+    indspre, indspost = indsA[1:dim-1], indsA[dim+1:end]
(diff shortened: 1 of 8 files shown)


timholy:
0.6 failure needs fixing in Rotations, presumably.

test/warp.jl
```diff
@@ -36,5 +36,25 @@ tfm = recenter(RotMatrix(-pi/4), center(img_pyramid))
 imgr = warp(img_pyramid, tfm)
 @test indices(imgr) == (0:6, 0:6)
 
-# I do this very complicated because turns out NaNs are hard to compare...
+# Use map and === because of the NaNs
```
Evizero: :)

src/resizing.jl
```diff
@@ -18,118 +18,144 @@ function restrict(A::AbstractArray, region::Union{Dims,Vector{Int}} = coords_spa
...
+    newinds = ntuple(i->i==dim?restrict_indices(indsA[dim]):indsA[i], Val{N})
+    out = similar(Array{typeof(first(A)/4+first(A)/2),N}, newinds)
     restrict!(out, A, dim)
     out
 end
 
 # out should have efficient linear indexing
-@generated function restrict!{T,N}(out::AbstractArray{T,N}, A::AbstractArray, dim)
+function restrict!{T,N}(out::AbstractArray{T,N}, A::AbstractArray, dim)
+    if dim > N
```
Evizero: Shouldn't we throw an error in such a case?

src/resizing.jl
```diff
@@ -18,118 +18,144 @@ function restrict(A::AbstractArray, region::Union{Dims,Vector{Int}} = coords_spa
...
-                    end
-                    out[indx+=1] = oneeighth*c+threeeighths*d
-                end
-            else
-                fill!(out, zero(T))
-                strd = stride(out, dim)
-                stride_1 = 1
-                @nexprs $N d->(stride_{d+1} = stride_d*size(out,d))
-                @nexprs $N d->offset_d = 0
+            z = convert(T, zero(first(A)))
```
Evizero: why not `z = zero(T)` ?

test/resizing.jl
```diff
@@ -30,6 +30,10 @@
     img1 = colorview(RGB, fill(0.9, 3, 5, 5))
     img2 = colorview(RGB, fill(N0f8(0.9), 3, 5, 5))
     @test isapprox(channelview(restrict(img1)), channelview(restrict(img2)), rtol=0.01)
+    # Non-1 indices
+    @test parent(restrict(OffsetArray(A, (-2,1,0)), 1)) == restrict(A, 1)
```
Evizero: did you consider:

```julia
@test parent(@inferred(restrict(OffsetArray(A, (-2,1,0)), 1))) == restrict(A, 1)
```

Evizero:
Very cool! Made some comments here and there when something caught my eye

src/resizing.jl
timholy: I'm not opposed to doing so, but I was mimicking this:
```julia
julia> a = 1:3
1:3

julia> sum(a, 1)
1-element Array{Int64,1}:
 6

julia> sum(a, 2)
3-element Array{Int64,1}:
 1
 2
 3
```

src/resizing.jl
timholy: Not certain, but I think I put it this way originally to support the following:
```julia
julia> a = [rand(3) for i = 1:5]
5-element Array{Array{Float64,1},1}:                                                                                                                                                         
 [0.465388,0.525425,0.753336]                                                                                                                                                                
 [0.890223,0.465896,0.191895]                                                                                                                                                                
 [0.205448,0.896783,0.0923702]                                                                                                                                                               
 [0.538556,0.862554,0.581898]                                                                                                                                                                
 [0.731725,0.887581,0.0116221]

julia> restrict(a, 1)
3-element Array{Array{Float64,1},1}:                                                                                                                                                         
 [0.45525,0.379187,0.424642]                                                                                                                                                                 
 [0.459918,0.780504,0.239633]                                                                                                                                                                
 [0.500502,0.659429,0.151285]
```
Clearly I should add that to the tests.

timholy:
Thanks for the review, @Evizero!

src/resizing.jl
Evizero: Ah, I see. I think consistency with Base is the cleanest way, yes

test/resizing.jl
timholy: In fact, there was a type instability due to [this method](https://github.com/JuliaImages/ImageTransformations.jl/blob/24912f9ea09b5ea7ff9285961fd1f0c6d0cc52de/src/resizing.jl#L13-L18), since `A` can change its type. Thanks for the suggestion!

Evizero:
Worth a tag? I am eyeing the merge of https://github.com/JuliaImages/ImageInTerminal.jl/pull/9

PR #20, opened 2017-04-12
Support indices for `imresize`

timholy:
Ref https://github.com/JuliaImages/ImageInTerminal.jl/pull/9#issuecomment-293495965

### DIFF
--- a/src/resizing.jl
+++ b/src/resizing.jl
@@ -156,18 +156,24 @@ end
 # imresize
 imresize(original::AbstractArray, dim1, dimN...) = imresize(original, (dim1,dimN...))
 
-function imresize{T,N}(original::AbstractArray{T,N}, short_size::NTuple)
+function imresize(original::AbstractArray, short_size::Tuple)
     len_short = length(short_size)
-    new_size = ntuple(i -> (i > len_short ? size(original,i) : short_size[i]), N)
+    len_short > ndims(original) && throw(DimensionMismatch("$short_size has too many dimensions for a $(ndims(original))-dimensional array"))
+    new_size = ntuple(i -> (i > len_short ? odims(original, i, short_size) : short_size[i]), ndims(original))
     imresize(original, new_size)
 end
+odims(original, i, short_size::Tuple{Integer,Vararg{Integer}}) = size(original, i)
+odims(original, i, short_size::Tuple{}) = indices(original, i)
+odims(original, i, short_size) = oftype(first(short_size), indices(original, i))
 
 """
     imresize(img, sz) -> imgr
+    imresize(img, inds) -> imgr
 
-Change `img` to be of size `sz`. This interpolates the values at
-sub-pixel locations. If you are shrinking the image, you risk aliasing
-unless you low-pass filter `img` first. For example:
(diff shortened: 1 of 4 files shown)


src/resizing.jl
```diff
@@ -156,34 +156,58 @@ end
 # imresize
 imresize(original::AbstractArray, dim1, dimN...) = imresize(original, (dim1,dimN...))
 
-function imresize{T,N}(original::AbstractArray{T,N}, short_size::NTuple)
+function imresize(original::AbstractArray, short_size::Tuple)
     len_short = length(short_size)
-    new_size = ntuple(i -> (i > len_short ? size(original,i) : short_size[i]), N)
+    new_size = ntuple(i -> (i > len_short ? odims(original, i, short_size) : short_size[i]), N)
```
Evizero: If `N` is removed from the type parameters, where does it come from? . How does this compile?

Evizero: Also I just now realize I could have written `Val{N}` in the original code, which I didn't for some reason

timholy: Must not be tested? But a missing var is detected at runtime, not compile time.

Evizero: Yes, apparently I added this method without adding corresponding tests. Sorry about that

timholy: No problem. I'm really glad you caught the error.

Evizero: If you don't mind I have a follow up question on "runtime vs compile time". I think I am mixing up terminology here and I would like to learn how to, well, avoid that.

- Based on my intuition I would assume that without the `N`, the JIT compilation would fail once the function is called. I.e. it could never produce any machine code that could be executed. (I understand, though, that in this particular example the function wasn't called because of the lack of any test).

- Since I still think the above is true (please do correct me if it isn't), I am guessing I am simply using the term "compile time" incorrectly.

Where do you, as a Julia developer, draw the line between compile time and runtime? Do you only count the execution of the Julia code that defines the function as compile time (like pasting a function in the REPL and pressing enter)? Is the JIT compilation thought of as a run-time event, and consequently if it fails it is called a runtime error?

timholy: Here's an illustrative example:
```julia
julia> foo() = n > 2
foo (generic function with 1 method)

julia> @code_warntype foo()
Variables:
  #self#::#foo

Body:
  begin 
      SSAValue(0) = Main.n
      return (2 < SSAValue(0))::Any
  end::Any
```
You can see that the inferred code will try to look up a variable `Main.n` when this code runs. If you look at this with `@code_llvm`, you'll see a block
```julia
notfound:                                         ; preds = %top
  %11 = call %jl_value_t* @jl_get_binding_or_error(%jl_value_t* inttoptr (i64 139624985559056 to %jl_value_t*), %jl_value_t* inttoptr (i64 139624983216904 to %jl_value_t*))
  store %jl_value_t* %11, %jl_value_t** @delayedvar70580, align 8
  br label %found
```
which, from the names alone, probably convinces you that the *compiled* code includes a *runtime* variable lookup. In other words, it compiles the lookup into the code; that's why it's a runtime failure.

Evizero: interesting example. that clears up a lot, thanks

PR #22, opened 2017-04-15
Ensure SVector construction preserves dimensionality of CartesianIndex

timholy:
`SVector` used to have a method so that `SVector(i::CartesianIndex)` created a vector of the components of `i`, but that method seems to be gone in StaticArrays 0.4. Since there are situations in which you'd want to create an `SVector` of `CartesianIndex`es, I'm not surprised this changed; it seems to be best to be explicit here about our intention.

### DIFF
--- a/src/autorange.jl
+++ b/src/autorange.jl
@@ -1,8 +1,8 @@
 function autorange(img, tform)
     R = CartesianRange(indices(img))
-    mn = mx = tform(SVector(first(R)))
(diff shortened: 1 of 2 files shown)


Evizero:
does it need a `REQUIRE` bump?

timholy:
I don't think it's necessary. Construction from an NTuple has always preserved N, it's only that the "convenience"/"magical" method is now gone.

PR #23, opened 2017-04-21
refactor warp to be consistent with WarpedView

Evizero:
Depends on #13. This follows the idea from `WarpedView` that the `eltype` of the input should be preserved, since the transformation should only be about warping.

### DIFF
--- a/src/warp.jl
+++ b/src/warp.jl
@@ -1,46 +1,24 @@
-# None of these types are provided by this package, so the following line seems unclean
-#     @inline Base.getindex(A::AbstractExtrapolation, v::StaticVector) = A[convert(Tuple, v)...]
-# furthermore it would be ambiguous with getindex(::Extrapolation, xs...) after https://github.com/tlycken/Interpolations.jl/pull/141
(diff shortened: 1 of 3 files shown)


src/warp.jl
```diff
@@ -1,46 +1,24 @@
...
-values are used to indicate locations for which `tform(x)` was outside
-the bounds of the input `img`. For more control over the interpolation
-scheme---and how beyond-the-edge points are handled---pass it in as an
-`AbstractExtrapolation` from Interpolations.jl.
+At off-grid points, `imgw` is calculated by interpolation. The
+default is linear interpolation, which is used when `img` is a
+plain array and the `img[tform(x)]` is inbound. In the case
+`tform(x)` maps to indices outside the original `img`, the value
+/ extrapolation scheme denoted by the optional parameter `fill`
+(which defaults to `NaN`) is used to indicate locations for which
```
timholy: Maybe specify `NaN`, if the element type supports it, and 0 otherwise?

Evizero:
closing in favour of #24

PR #24, opened 2017-04-21
Streamline API for warp and WarpedView

Evizero:
With this API refactor the following is guaranteed to hold in all cases (well I still need to write the tests, but that is the basic idea)

```julia
wv = WarpedView(A, tfm)
parent(wv) === A
```

I introduce a new function `warpedview` which allows a little convenience concerning the interpolation/extrapolation settings. It will still enforce "view" property (so no `Quadratic` etc), but it will make the API nicer to work with. 

```julia
wv = warpedview(A, tform, Constant())
wv = warpedview(A, tform, Linear())
wv = warpedview(A, tform, 1) # just extrapolation fill value
wv = warpedview(A, tform, Linear(), 1)
```

The functions above will cause `parent(wv) !== A`, because they are convenience functions that simulate a user calling `WarpedView(extrapolate(interpolate(A,...),...), tform)`. This is why these signatures aren't part of `WarpedView`. I think this makes the design cleaner. Also with this change, the summary string is now always accurate and no longer based on guessing.

Note I have some follow up plans that would expand `warpedview` further, so this isn't the only reason for this function to exist.

absorbs #23 

tests are currently broken because I did not update them yet


Evizero:
Sorry that my PRs tend to be so difficult to review in this package. Since the package is quite young, I am still moving code around if it things look out of place. Better now than later

timholy:
I like the `WarpedView` guarantee at top. I also like `warpedview`, and on balance it may be the right API. However, the obvious concern is that despite its name it doesn't create a view of the input.

A relevant comparison is `colorview` vs `ColorView`. Both of these create a view, but the distinction is that the return type of `ColorView` is `ColorView`, whereas `colorview` might return any suitable view type that "acts like" `ColorView`. (e.g., it will call `reinterpret` when appropriate.) I worry a bit that introducing an uppercase/lowercase pair with different behavior might be a bit confusing.

The main alternative that occurs to me is to introduce a `warp_prepare` or `warp_extrapolant` function that focuses on the preparation of the `etp` and then pass that as an argument to `WarpedView`. I'm not entirely thrilled with this, either, so this is firmly in "discussion" mode and not meant to be a TODO item.

Evizero:
> However, the obvious concern is that despite its name it doesn't create a view of the input.

Can you elaborate on that? Why is it not a view? Because of the interpolation?
Other than that my current mental model of what makes something a view seems satisfied. 

screen shot 2017-04-21 at 20 11 43

Evizero:
or are you explicitly taking about the lower case version ?

timholy:
Perhaps I was confused. I assumed you meant you would support `warpedview(A, tform, Quadratic(Flat()))`. Now that I suspect the answer is "no," then I rescind my complaint. :+1:

Evizero:
Oh, understood. No, I had the same concerns actually.

For `WarpedView` I do not allow convenience parameters at all. But is it possible to supply a prepared interpolation/extrapolation.

For `warpedview`, I restrict the parameters for the interpolation to `Linear()` or `Constant()` (the `BSpline(...)` part is assumed and hard-coded). I also create the `BSplineInterpolation` manually to prevent `prefilter`. The extrapolation can be numbers, colorants, or flags such as `Flat()` and `Periodic()`. If one calls `warpedview` with custom settings, it will act as one called `WarpedView` with a manually prepared extrapolation that is still a view around the original array. This fact is also accurately displayed in the summary string.

screen shot 2017-04-21 at 20 25 37

Naturally it is allowed to manually pass *any*  interpolation or extrapolation to both functions. But then it was a user decision.

src/warpedview.jl
```diff
@@ -1,62 +1,86 @@
...
+    itp = Interpolations.BSplineInterpolation{T,N,typeof(parent),BSpline{D},OnGrid,0}(parent)
+    _build_extrapolation(itp, args...)
+end
+
+immutable WarpedView{T,N,A<:AbstractArray,F1<:Transformation,I,F2<:Transformation,E<:AbstractExtrapolation} <: AbstractArray{T,N}
     parent::A
     transform::F1
-    transform_inv::F2
     indices::I
+    inverse::F2
```
Evizero: I did notice the `inv` part quite early. Well, that sounds more observant as it is, because basically I had to reverse engineer what the code does in order to understand what is going on. I also spend some thoughts on the issue you describe.

I like this approach for the simple reason as it behaves how I want it to behave. I like to reason about how the image changes, not how the indices change. Also I do want the option to nest `WarpedView` for a use case I have. Basically I like this way because it is convenient for me.

Why don't we avoid commitment to one interpretation altogether. We could add a `InvWarpedView` (and `invwarp`)  which does `w = img[tfm(x)]` as an alternative.

timholy: That sounds like the sanest approach. We should probably document somewhere that the transformation is effectively applied to the image rather than the coordinates. It's probably more mathematically general/elegant to have it apply to the coordinates, but I agree that intuition goes the other way. And consistency with other languages (assuming Matlab isn't the only one) is a good thing.

Evizero: We do also have a choice which one we call `InvWarpedView`

Evizero: In the `R` package `imager` it seems that a user has the choice to specify `"forward"` or `"reverse"` mode, where forward (which is ironically what we do right now with `inv`) is the default

Evizero: Maybe `FwdWarpedView` and `BwdWarpedView`? Although I do think that one of them should just be `WarpedView`. Personally I vote for keeping the current one as is, and adding the alternative with one of the following names (any other ideas?)

- `InvWarpedView`
- `BwdWarpedView`
- `IdsWarpedView`
- `WarpedIndicesView`

timholy: Sure, let's keep the current one `WarpedView`. I like `WarpedIndicesView`, could also consider`WarpedCoordsView`.

Evizero:
Ok, so now I kinda changed my mind again. So here are my thoughts on the `inv` issue.

First I thought we are kinda performing forward mode warping, since be supply the forward transformation. However, I have come to appreciate that, no, we are not. We are performing backward warping with an - for that functionality - unintuitive interface. We do backward warping by specifying the forward transform.

So really I think we should change how `warp` itself works to "backward" mode, which means that it should do `w = img[tfm(x)]` instead of using `inv`. Actually I tend to think we should additionally add some parameter for dispatch `warp(A, tfm, WarpMode.Backward())`. This would allow us to add `Forward()` in the future.

We can change `WarpedView` to then have a dispatch-able parameter for `Backward` as well.

The functionality I am after right now with the *current* `WarpedView` could be renamed to `InvWarpedView`, which I'd be fine with.

Evizero:
sorry for causing this kind of confusions. Since I am leaning out of my field of expertise I often have to study up on topics/related work as I go along or issues emerge. My background is ML and not image processing specifically. Thanks for all that feedback. These quick responses you keep giving me are really appreciated.

src/warp.jl
```diff
@@ -70,39 +94,26 @@ julia> indices(imgr)
 (Base.OneTo(906),Base.OneTo(905))
 ```
 """
-function warp{T}(img::AbstractExtrapolation{T}, tform)
-    inds = autorange(img, tform)
+function warp{T}(img::AbstractExtrapolation{T}, tform, inds::Tuple = autorange(img, inv(tform)))
```
Evizero: we still need to use `inv` to compute the indices. But I changed it so that a user can provide them instead and thus avoid `inv`

Evizero:
I ported and adapted the ImageInTerminal style of comparing to small reference images (each takes around 5-15 kb). For warping these small thumbnail images are actually informative enough for most tests, so I thought this will make it a lot easier to really test the API properly.

The one downside of this is that the reference files aren't really that useful to look up on github.

andyferris:
The forward and backward thing is interesting. This seems to also be a problem in CoordinateTransformations, where it's not really clear whether a `Transformation` is active or passive. 

In some cases like changing coordinate systems it is passive (the point doesn't move - it's expressed in a new frame). However, it seems things like `LinearMap` are more designed as active transformations e.g. you want to rotate the points about an axis, say. 

My point being that if we find a way of expressing the difference upstream in CoordinateTransformations, perhaps it would solve the ambiguity here. Cc @c42f

timholy:
To avoid polluting this PR, I've opened #25 to discuss the forward/backward thing.

Evizero:
I feel bad about triggering all the codecov spam. Any way we can reduce these messages?

Evizero:
latest tests require https://github.com/JuliaMath/Interpolations.jl/pull/150

Evizero:
How about this? This is non breaking. After a little deprecation period we could reassign the new meaning to `warp(img, tform)`

Evizero:
Or as suggested here: https://github.com/JuliaImages/ImageTransformations.jl/issues/25#issuecomment-296524277, we could also just rip off the band-aide right away and skip deprecation. Just wanted to put an option on the table

c42f:
Do you plan to support `Backward` and `Forward` in the longer term?  If not, isn't this deprecation strategy functionally equivalent to

```julia
warp_new(img::AbstractArray, inv_tform, args...) = # ... implementation
@deprecate warp(img::AbstractArray, tform, args...) warp_new(img, inv(tform), args...)
```

Followed by, after the deprecation period
```julia
warp(img::AbstractArray, inv_tform, args...) = # ... implementation
@deprecate warp_new(img::AbstractArray, inv_tform, args...) warp(img, inv_tform, args...)
```

That is, there's no real need to bundle it up into a single generic function `warp` with a trait for dispatch - you can just have two names for the function.  In both cases, it's a two stage deprecation cycle where users have to change things their code twice.

Evizero:
> Do you plan to support Backward and Forward in the longer term?

Well that is what some other frameworks seem to do https://dahtah.github.io/imager/imager.html#warping

Evizero:
That said, we can always introduce forward mode at some later point. 

Let's go with your simpler suggestion for now then.

c42f:
I'm not against forward mode at all, as long as there's a good motivation for needing it.

However, from the imager page:

> In forward mode you go through all (x,y) pixels in the source, and paint the corresponding location M(x,y) in the target image. This may result in unpainted pixels, as in the following example

To the extent that a pixelized image is usually a discrete representation of a continuous function, "unpainted pixels" seem somewhat fundamentally wrong to me, and makes me question whether the design choices from imager are particularly solid.

For warping, I prefer the conceptual formulation
discrete representation (pixels) -> continuous reconstruction -> continuous warp -> antialiasing filter -> discretize

I remember greatly enjoying Paul Heckbert's masters thesis on this topic - https://www.cs.cmu.edu/~ph/texfund/texfund.pdf

c42f:
(BTW, keep in mind that my background in this stuff is from the point of view of high quality image resampling for computer graphics, so while I know a fair bit about how to write fast pointwise texture samplers with correct antialiasing, I'm not a domain expert in scientific image processing.)

Evizero:
What you say makes sense to me. 

Really, at this point I think we should just make a decision. I'd say let's do the `warp_new` deprecation strategy so we can make the transition sooner rather than later. Working on the commit now.

timholy:
The `warp_new` strategy is much like option 2 in https://github.com/JuliaImages/ImageTransformations.jl/issues/25#issuecomment-296442100. If we want to minimize the code churn for users, should we do that variant "completely"?

Evizero:
How about this?

src/warp.jl
```diff
@@ -1,51 +1,35 @@
...
 """
-    warp(img, tform) -> imgw
+    warp(img, tform, [indices], [degree = Linear()], [fill = NaN]) -> imgw
 
 Transform the coordinates of `img`, returning a new `imgw` satisfying
-`imgw[x] = img[tform(x)]`. `tform` should be defined using
-CoordinateTransformations.jl.
+`imgw[I] = img[tform(I)]`. This approach is known as backward
+mode warping. The transformation `tform` should be defined using
+[CoordinateTransformations.jl](https://github.com/FugroRoames/CoordinateTransformations.jl).
```
timholy: Not strictly required that it come from CoordinateTransformations, is it? Anything that can take an `SVector` input should work?

src/warp.jl
```diff
@@ -1,51 +1,35 @@
...
-At off-grid points, `imgw` is calculated by interpolation. The default
-is linear interpolation, used when `img` is a plain array, and `NaN`
-values are used to indicate locations for which `tform(x)` was outside
-the bounds of the input `img`. For more control over the interpolation
-scheme---and how beyond-the-edge points are handled---pass it in as an
-`AbstractExtrapolation` from Interpolations.jl.
+At off-grid points, `imgw` is calculated by interpolation. The
+degree of the b-spline can be specified with the optional
+parameter `degree`, which can take the values `Linear()` or
+`Constant()`.
```
timholy: This raises an interesting issue: since `warp` doesn't have a "view" pledge to keep, could it support `Quadratic` or `Cubic` directly?

src/warp.jl
```diff
@@ -1,51 +1,35 @@
...
+
+The b-spline interpolation is used when `img` is a plain array
+and `img[tform(I)]` is inbound. In the case `tform(I)` maps
+to indices outside the original `img`, the value or extrapolation
```
timholy: might be clearer to say "those locations are set to a value `fill` (which defaults...)" and delete the material after the closing parenthesis.

src/warp.jl
```diff
@@ -1,51 +1,35 @@
...
+scheme denoted by the optional parameter `fill` (which defaults
+to `NaN` if the element type supports it, and `0` otherwise) is
+used to indicate locations for which `tform(I)` was outside the
+bounds of the input `img`.
+
+For more control over the interpolation scheme --- and how
+beyond-the-edge points are handled --- pass it in as an
```
timholy: Maybe "pass `img` in as..."?

src/warp.jl
```diff
@@ -65,7 +49,7 @@ julia> indices(img)
 julia> tfm = recenter(RotMatrix(pi/4), center(img))
 AffineMap([0.707107 -0.707107; 0.707107 0.707107], [347.01,-68.7554])
 
-julia> imgw = warp(img, tfm);
+julia> imgw = warp(img, inv(tfm));
```
timholy: I think it would be better to pass `tfm` directly. You could change the angle to `-pi/4` above.

src/warp.jl
```diff
@@ -87,28 +71,117 @@ julia> indices(imgr)
 (Base.OneTo(906),Base.OneTo(905))
 ```
 """
-function warp(img::AbstractExtrapolation, tform)
-    inds = autorange(img, tform)
-    out = OffsetArray(Array{eltype(img)}(map(length, inds)), inds)
+function warp_new{T}(img::AbstractExtrapolation{T}, tform, inds::Tuple = autorange(img, inv(tform)))
```
timholy: Someday this will have to become `tform::InvertibleTransformation` or similar.

src/warp.jl
```diff
@@ -87,28 +71,117 @@ julia> indices(imgr)
...
+    out = OffsetArray(Array{T}(map(length, inds)), inds)
     warp!(out, img, tform)
 end
 
+# this function was never exported, so no need to deprecate
```
timholy: Agreed.

src/warp.jl
```diff
@@ -87,28 +71,117 @@ julia> indices(imgr)
...
+# # after deprecation period:
+# @deprecate warp_new(img::AbstractArray, tform, args...) warp(img, tform, args...)
+# @deprecate warp_old(img::AbstractArray, tform, args...) warp(img, tform, args...)
+
+"""
+    warp(img, tform, [indices], [degree = Linear()], [fill = NaN]) -> imgw
+
+!!! warning "Deprecation Warning!"
+
+    This method with the signature `warp(img, tform, args...)` is
```
timholy: I'm not sure we need to document the `warp_old` in as much detail, perhaps only enough to get people to switch:
```
    `warp` is transitioning to a different interpretation of the transformation, and you are using
    the old version.
    Please see https://github.com/JuliaImages/ImageTransformations.jl for information about
    how to switch to the new version.
```
Or we can specify how one switches in the docstring (e.g., to help people who run into this at a time when they don't happen to have internet access, e.g., while flying).

src/warp.jl
```diff
@@ -87,28 +71,117 @@ julia> indices(imgr)
...
+julia> indices(imgw)
+(-293:612,-293:611)
+
+julia> imgr = parent(imgw);
+
+julia> indices(imgr)
+(Base.OneTo(906),Base.OneTo(905))
+```
+"""
+@generated function warp_old(img::AbstractArray, tform, args...)
```
timholy: Does this need to be a `@generated` function? Can't it just call `warp_new` directly?

src/warp.jl
```diff
@@ -87,28 +71,117 @@ julia> indices(imgr)
...
+    warn("'warp(img, tform)' is deprecated in favour of the cleaner interpretation 'warp(img, inv(tform))'. Set 'const warp = ImageTransformations.warp_new' right after package import to change to the new behaviour right away.")
```
timholy: Shall we use the `Base.depwarn` mechanism? And perhaps rather than call it "cleaner" (which is a value judgement), just refer people to the docstring and/or the web site?

The web site can provide more justification, e.g., like my post in https://github.com/JuliaImages/ImageTransformations.jl/issues/25#issue-223569275 except probably using embedded pictures.

test/deprecations.jl
```diff
@@ -0,0 +1,7 @@
+info("Start of deprecation warnings")
+
+img_camera = testimage("camera")
+tfm = recenter(RotMatrix(-pi/8), center(img_camera))
+imgr = @inferred(ImageTransformations.warp_old(img_camera, tfm))
+@test eltype(imgr) == eltype(img_camera)
+@test_reference "warp_cameraman_rotate_r22deg" imgr
```
timholy: Thanks for going to the effort to test that the deprecation works!

test/runtests.jl
```diff
@@ -1,15 +1,74 @@
...
+                    write(reference_path(filename), join(actual, "\n"))
+                end
+            else
+                error("You need to run the tests interactively with 'include(\"test/runtests.jl\")' to create new reference images")
+            end
+        else
+            throw(ex)
+        end
+    end
+end
```
timholy: Beautiful! Such a nice addition.

test/warp.jl
```diff
@@ -1,53 +1,49 @@
...
 
 img_camera = testimage("camera")
 @testset "Constructor" begin
     tfm = recenter(RotMatrix(-pi/8), center(img_camera))
     ref_inds = (-78:591, -78:591)
 
-    @testset "warp" begin
-        imgr = @inferred(warp(img_camera, tfm))
+    @testset "warp_new" begin
+        imgr = @inferred(warp(img_camera, inv(tfm)))
```
timholy: I'd rather define `tfm` above with `RotMatrix(pi/8)` and not call `inv`.

src/invwarpedview.jl
```diff
@@ -0,0 +1,74 @@
+immutable InvWarpedView{T,N,A<:AbstractArray,F1<:Transformation,I,F2<:Transformation,E<:AbstractExtrapolation} <: AbstractArray{T,N}
```
Evizero: good idea. I will think on that

src/interpolations.jl
```diff
@@ -0,0 +1,49 @@
...
+
+# The default values used by extrapolation for off-domain points
+@compat const FillType = Union{Number,Colorant,Flat,Periodic,Reflect}
+@compat const FloatLike{T<:AbstractFloat} = Union{T,AbstractGray{T}}
+@compat const FloatColorant{T<:AbstractFloat} = Colorant{T}
+@inline _default_fill{T<:FloatLike}(::Type{T}) = convert(T, NaN)
+@inline _default_fill{T<:FloatColorant}(::Type{T}) = _nan(T)
+@inline _default_fill{T}(::Type{T}) = zero(T)
+
+box_extrapolation(etp::AbstractExtrapolation) = etp
```
Evizero: Just that the image is "boxed" or "decorated" by an extrapolation. It's supposed to signify that this will be a view.

Can you think of a better name?

src/warp.jl
Evizero: I'll rewrite it to be more neutral

src/warp.jl
Evizero: I was aiming for for consistency with the view.

Regardless of consistency, I do think that we shouldn't hide such allocating actions from a user. It should be a choice. It isn't that much more effort to pass `interpolate(img, ...)` to the constructor.

The options presented are special in that they are quite conservative in the memory allocation regard.

src/warp.jl
Evizero: Good point

src/warp.jl
Evizero: The reason for `@generated` is to only present the warning once.

src/warp.jl
Evizero: Not sure what you are referring to with the comment on embedded pictures.

src/warp.jl
```diff
@@ -1,51 +1,35 @@
...
 
 # Interpolation scheme
 
-At off-grid points, `imgw` is calculated by interpolation. The default
-is linear interpolation, used when `img` is a plain array, and `NaN`
-values are used to indicate locations for which `tform(x)` was outside
-the bounds of the input `img`. For more control over the interpolation
-scheme---and how beyond-the-edge points are handled---pass it in as an
-`AbstractExtrapolation` from Interpolations.jl.
+At off-grid points, `imgw` is calculated by interpolation. The
```
c42f: I suggest abandoning the idea that "on grid points" are particularly meaningful.  This is only an artifact of certain interpolation schemes.

From a camera, a pixel value is not a gridded *point sample* of some underlying continuous image.  (Neither is it a little square "pixel box" we've all grown so used to looking at - pixel boxes are just the zeroth order b-spline interpolant.) In reality, the image is the incoming light field convolved with the point spread function of the optics, and integrated over the physical CCD/cmos sensor pixel.  The pixels are then a discrete low dimensional representation of the underlying light field.

All of which is to say... I'm not sure what!  Perhaps just that we shouldn't confuse the standard (ie, cheap) interpolation schemes with the interpretation of the pixels values.

Evizero: Alternatives:

- "At off-index points ..."

- "At points between indices ..."

Other ideas?

src/warp.jl
c42f: I agree with Tim here, I think it would be very natural to support other interpolation types.  (Caveat: I don't think I currently have a good overview of what you're building in this package as a whole.)

src/warp.jl
```diff
@@ -1,51 +1,35 @@
...
+to indices outside the original `img`, the value or extrapolation
+scheme denoted by the optional parameter `fill` (which defaults
+to `NaN` if the element type supports it, and `0` otherwise) is
+used to indicate locations for which `tform(I)` was outside the
+bounds of the input `img`.
+
+For more control over the interpolation scheme --- and how
+beyond-the-edge points are handled --- pass it in as an
+`AbstractInterpolation` or `AbstractExtrapolation` from
+[Interpolations.jl](https://github.com/JuliaMath/Interpolations.jl).
```
c42f: It's interesting how these packages fit together.

It might be worth noting that `interpolations` isn't the only piece of the puzzle regarding image warping.  In particular, it's never going to deal with proper antialiasing which can be quite important when the warping is severe or you're downsampling by a large factor.

For severe warping and large downsampling things can get quite interesting, as you need full anisotropic filtering, usually coupled with image pyramiding to make it efficient.  Arguably, `warp` could do these things with appropriate options turned on.

src/warp.jl
c42f: Presumably invertibility is more of a trait?  I can't see it fitting neatly into the type hierarchy.

test/reference/invwarpedview_cameraman.txt
```diff
@@ -0,0 +1,17 @@
+[0m[38;5;236;48;5;239m▀[38;5;239;48;5;246m▀[38;5;239;48;5;246m▀[38;5;239;48;5;246m▀[38;5;239;48;5;247m▀[38;5;239;48;5;247m▀[38;5;240;48;5;247m▀[38;5;240;48;5;247m▀[38;5;240;48;5;248m▀[38;5;240;48;5;248m▀[38;5;240;48;5;248m▀[38;5;240;48;5;248m▀[38;5;240;48;5;248m▀[38;5;240;48;5;249m▀[38;5;240;48;5;249m▀[38;5;240;48;5;249m▀[38;5;240;48;5;249m▀[38;5;240;48;5;249m▀[38;5;240;48;5;248m▀[38;5;240;48;5;248m▀[38;5;240;48;5;248m▀[38;5;240;48;5;248m▀[38;5;240;48;5;248m▀[38;5;240;48;5;247m▀[38;5;240;48;5;247m▀[38;5;240;48;5;247m▀[38;5;240;48;5;247m▀[38;5;239;48;5;247m▀[38;5;239;48;5;246m▀[38;5;239;48;5;246m▀[38;5;239;48;5;246m▀[38;5;239;48;5;246m▀[38;5;235;48;5;239m▀[0m
```
c42f: Oh ho.  This is very cool.

src/warp.jl
c42f: (Obviously, this is a side comment - nothing to do with any changes you should be making here in this PR.)

Evizero: We currently have the same kind of problem with `imresize`

see https://github.com/JuliaImages/ImageTransformations.jl/pull/7#issue-208597726

src/warp.jl
c42f: It sounds like I might be missing something about the way images are treated in this package, or about the way that interpolations works in conjunction with this.

Normally (ie, in all other software packages I know about), cubic spline image resampling technically gives an approximating spline rather than an interpolating one. Is that not the case here?

src/warp.jl
```diff
@@ -37,7 +36,7 @@ with `parent(imgw)`.
 
 # Examples: a 2d rotation (see JuliaImages documentation for pictures)
 
-```jldoctest
```
Evizero: doc tests wouldn't pass given that the following isn't the "current" behaviour

Evizero:
Open points / debates before merging

- [ ] Allow `Quadratic` and `Cubic` to be passed to `warp` directly? Can we do a vote on that? 

- [ ] Grid discussion https://github.com/JuliaImages/ImageTransformations.jl/pull/24#discussion_r113672501 (hoping for some reply from @timholy on that one)

- [ ] Some more tests for the different `warp` signatures (I know which ones, just need to implement them).

- [ ] docstring for `InvWarpedView` and `invwarpedview`

- [ ] (maybe another PR) Add `WarpedView` and use as basis for `InvWarpedView`

anything else?

src/warp.jl
timholy: For quadratic and cubic, Interpolations solves a tridiagonal/Woodbury system for the coefficients to ensure that they are interpolating.

src/warp.jl
timholy: Hmm, yes, agreed.

src/warp.jl
timholy: To address @c42f's point more completely, yes, agreed 100% with your physical model of the process. The issue is that without more information there isn't, to my knowledge, a great way to do a lot better than interpolation. So personally I'm fine with talking about a "grid," though I'd also be fine with a slightly more verbose statement "if `tform(I)` does not have integer coordinate values, the image value is computed by interpolation."

timholy:
I am in favor of the "view" guarantee if you're explicitly creating a view via, e.g., `invwarpedview`. But `warp` doesn't promise a view, and that's where I think it's reasonable to support non-view interpolation schemes. As you say, I don't think this is a make-or-break point, since it's not hard to create the interpolation object directly. So I'll be fine however this decision comes out.

timholy:
>  I think looking at the same warning every time one calls the method might be a bit of an overkill.

`Base.depwarn` collects a stacktrace and then doesn't print the warning if it's from the same originating line of code. So (when the line numbering works, which it doesn't always), you get exactly one depwarn per executed line of code that needs changing, even if you're looping over something.

src/warp.jl
c42f: Thanks for clearing that up.  I'm also not sure what more should be done, other than some kind of interpolating or approximating spline.  (Or at least, that anything else which could be done is far too fancy to be done behind the user's back.)  This makes more sense of @Evizero's worry about using higher-than-linear orders inside `warp`.

I must admit it's not *quite* clear to me that an interpolating spline is the right choice in general over the much simpler and - at least somewhat - more efficient approximating splines for image resampling.  In the alternative kernel-based view of reconstructing a continuous image from discrete pixel values, there's a convenient family of bicubic kernels parameterizing the apparent sharpness of the resulting reconstruction.  This formulation makes the reconstruction entirely spatially local, and there's no need to run any kind of solver before sampling in an arbitrary location.

Perhaps all I'm quibbling about is that the docs give the idea that the computation is different on vs off the integer lattice.  It shouldn't be, other than as an optimization in the case that the reconstruction is done with an actual interpolating spline.

timholy: > there's a convenient family of bicubic kernels parameterizing the apparent sharpness of the resulting reconstruction

@c42f, a reference would be awesome.

> the docs give the idea that the computation is different on vs off the integer lattice. It shouldn't be

That's absolutely right. Do you have a good way to phrase this? I wondered if that's what you were worried about, but I couldn't quickly come up with a concise way of expressing that.

c42f: It's mentioned in passing on the wikipedia page about bicubic interpolation, but it's not great. I finally found a better reference in some old code of mine:

> Filter kernels for bicubic reconstruction filters are discussed in detail in
> "Reconstruction filters in computer graphics", D. Mitchell and A. Netravali,
> Computer Graphics, vol. 22 no. 4, 1988.
>
> Mitchell and Netravali identify two desirable properties for signal
> reconstruction:
>
>   * We want the image and its first derivative to be continuous
>   * We want constant areas of an image to stay constant after interpolation
>
> This narrows down the possible filters to a two parameter space; parameters
> B and C in the paper.  In addition, restricting to the line 2C + B = 1
> ensures quadratic convergence as the pixel size is reduced; we use a
> qualitative parameter "sharpness = 1-B" below.
>
> For a reasonably sharp filter, try sharpness = 1; this is the classic
> Catmull-Rom spline.  For an image sharpness _enhancing_ filter, try
> sharpness = 2.
>
> A somewhat blurry result (but without any ringing) will be obtained using
> the cubic B-spline, which corresponds to sharpness = 0.

The code itself was just an implementation of the piecewise formulae from the paper.

c42f: > Do you have a good way to phrase this

Not exactly :-) How about

```
# Reconstruction scheme

During warping, values for `img` must be reconstructed at arbitrary
locations `tform(I)` which do not lie on to the lattice of pixels. How
this reconstruction is done depends on the type of `img` and the
optional parameter `degree`.

When `img` is a plain array, b-spline interpolation is used with
`degree=Linear()` for linear interpolation or `degree=Constant()` for
nearest neighbor interpolation.  In the case ...
```

Obviously I'm trying to avoid the word "interpolation", as there's quite reasonable non-interpolating reconstructions.

Evizero:
good feedback. will work it in

src/warp.jl
c42f: Having had the long discussion about "off-grid points" I think I now understand the context for your reservations about higher orders, and the view pledge which Tim is referring to.

But I also think we could do a cubic spline view just fine, as long as we take a convolution kernel-style viewpoint of the reconstruction (interpolation/approximation) process: in this variant, there's no matrix solver to run after modifying the underlying image, you only need the pixel values from a local 4x4 image patch around the sample point, and the fractional coordinates of the sample point away from the integer lattice.

src/warp.jl
c42f: Musing a bit more about reconstructions and the desiderata from the Mitchell paper -
* The Catmull-Rom kernel referred to above gives an interpolating spline, IIRC.  Other values of sharpness give approximating splines.
* Natural images are composed of discontinuous geometric edges and smooth lighting variations. Reconstructing such an image with a smooth spline is practical and looks reasonably good, but can't ever be quite the right thing for the non-smooth parts of an image. (Ultimately I guess the "right thing" is something more like a statistical model such as https://arxiv.org/abs/1702.00783 :-)  Not that I'm remotely suggesting doing this, but it's interesting to have in mind as a point of comparison with what we're actually doing.)

src/warp.jl
timholy: Sure. But we don't have that implemented yet, so let's finish this so that it works for what we have.

I can go either way (restrict to `Constant` and `Linear`, or support higher orders).

c42f: Sure, there's lots of great stuff here, I don't mean to hold up getting it merged.

Evizero:
I think the code is in place now. Just need to verify that with some tests and write some docstrings

Evizero:
One discussion point maybe. We tend to call the transformation variables `tform` or `tinv`. Should be consider naming them `passive` and `active` instead?

c42f:
I prefer tform and tinv over the words active and passive. I think the terms are uncommon enough and the concept subtle enough that users will have to look at the documentation anyway. And if they do, it's going to be easier to discuss in the documentation with words like `tform` which are more of an obvious placeholder for the object than a normal word like `active.

src/warp.jl
```diff
@@ -74,10 +74,13 @@ julia> indices(imgr)
 ```
 """
 function warp_new{T}(img::AbstractExtrapolation{T}, tform, inds::Tuple = autorange(img, inv(tform)))
-    out = OffsetArray(Array{T}(map(length, inds)), inds)
+    out = _allocate_array(T, inds)
     warp!(out, img, tform)
 end
 
+_allocate_array{T,N}(::Type{T}, inds::NTuple{N,Base.OneTo}) = Array{T}(map(length, inds))
+_allocate_array{T,N}(::Type{T}, inds::NTuple{N,AbstractUnitRange}) = OffsetArray(Array{T}(map(length, inds)), inds)
```
Evizero: unsure if this change could be considered controversial, but I did it because now behaviour should is consistent between `warp(...)` and `collect(warpedview(...))`. It does negatively affect type inference

```julia
julia> summary(warp(img_camera, tfm, indices(img_camera)))
"512×512 Array{Gray{N0f8},2}"

julia> summary(collect(warpedview(img_camera, tfm, indices(img_camera))))
"512×512 Array{Gray{N0f8},2}"
```

and 

```julia
julia> summary(warp(img_camera, tfm))
"-78:591×-78:591 OffsetArray{Gray{N0f8},2}"

julia> summary(collect(warpedview(img_camera, tfm)))
"-78:591×-78:591 OffsetArray{Gray{N0f8},2}"
```

Evizero:
Any ideas where this error on travis for `nightly` suddenly comes from?

```
...
Test Failed
  Expression: isempty(setdiff(ambs, refambs))
ERROR: LoadError: There was an error during testing
while loading /home/travis/.julia/v0.6/ImageTransformations/test/runtests.jl, in expression starting on line 7
...
```

Evizero:
Ready for the last round of reviews

src/interpolations.jl
timholy: I was tempted by `wrap_extrapolation`, but I think that could be confusing: people might think it's a type of `warp`! So let's stay with `box`.

src/invwarpedview.jl
```diff
@@ -0,0 +1,118 @@
+"""
+    InvWarpedView(img, tinv, [indices]) -> wv
+
+Create a view of `img` that lazily transforms any given index `I`
+passed to `wv[I]` to correspond to `img[inv(tinv)(I)]`. While
+technically this approach is known as backward mode warping,
+`InvWarpedView` is created using the forward transformation.
```
timholy: Perhaps "note that `InvWarpedView` is created by supplying the forward transformation"?

I wonder if it's a little confusing to call this parameter `tinv`, and then talk about `inv(tinv)`? I guess the alternative is to just call it `tform`. I recognize that this has the problem of having `tform` mean different things for `WarpedView` and `InvWarpedView`.

src/warp.jl
timholy: The behavior of `collect` is changing in 0.6, see https://github.com/JuliaLang/julia/pull/21257. In particular we decided that `collect` was the right way to spell "throw away the custom indices and return an Array with the same values." (See https://github.com/JuliaArrays/IdentityRanges.jl/issues/1.)

Nevertheless I like this change overall. I would write the second, however, as `similar(Array{T}, inds)`. That way the indices are in charge of the output type. When `inds` are UnitRanges, then it will allocate an OffsetArray.

src/invwarpedview.jl
Evizero: My personal justification for using `tinv` is that it is the inverse of the transformation that is actually requested by the algorithm (which is backward mode warping). I agree it doesn't look nice, but at least it gives it some consistency

If you feel like `InvWarpedView` might be a bit too far out there for the context of this package I could also just move it to the package where I actually need it,

src/warp.jl
Evizero: oh, yeah. that is a lot nicer. thanks

timholy:
If I don't notice, feel free to merge once this gets through CI. Thanks for doing this! A *huge* step forward.

Evizero:
This should be a squash merge since I was very casual with my commits

src/invwarpedview.jl
timholy: I'm fine either way, whichever you think will be best.

Evizero: Lets keep it here then. I like this additional layer of quality control. I also suspect that between `InvWarpedView` and `WarpedView`, the former is more convenient for a user to work with

Evizero:
Thanks for all the feedback and discussions. This was a fun PR!

c42f:
Sorry I didn't have more time to put into reviewing this, I don't know how Tim does it!  Great effort btw.

PR #26, opened 2017-05-03
Special treatment for SubArray of InvWarpedView

Evizero:
Context: For a use-case of mine I'd like to lazily be able to 1.) chain affine transformations, and 2.) allow for cropping in-between these operations. While 1. is already possible with `InvWarpedView` (after the merge of #24 ), point 2. is a little inconvenient right now, as it requires nesting `SubArray` and `InvWarpedView` over and over.

With this PR the package would employ special treatment for `invwarpedview(::SubArray{T,N,<:InvWarpedView}, ...)` in which it makes sure that only one `SubArray` of a single `InvWarpedView` is used.

```julia
julia> using ImageTransformations, CoordinateTransformations, Rotations, ImageInTerminal, TestImages, StaticArrays

julia> img_camera = testimage("camera");

julia> tfm = recenter(RotMatrix(-pi/8), center(img_camera))
AffineMap([0.92388 0.382683; -0.382683 0.92388], [-78.6334,117.683])

julia> iwv = invwarpedview(img_camera, tfm)
```

screen shot 2017-05-03 at 16 01 00

```julia
julia> v = view(iwv, 1:512, -70:400)
```

screen shot 2017-05-03 at 16 01 52

```julia
julia> tfm2 = LinearMap(SMatrix{2,2}([1 0; 0 .5]))
LinearMap([1.0 0.0; 0.0 0.5])

julia> v2 = invwarpedview(v, tfm2)
```

screen shot 2017-05-03 at 16 02 31

Here the last result (`v2`) looks equivalent to the result of `invwarpedview(collect(v), tfm2)`.
Naturally that is in general not the case, as `collect` throws away information that is otherwise secretly preserved underneath. But I don't think that keeping this information is a downside. Here is an example for what I mean:

```julia
julia> tfm3 = LinearMap(SMatrix{2,2}([1 0; .5 1]))
LinearMap([1.0 0.0; 0.5 1.0])

julia> v3 = invwarpedview(v2, tfm3)
``` 

screen shot 2017-05-03 at 16 07 42

```julia
julia> v4 = invwarpedview(collect(v2), tfm3)
``` 

screen shot 2017-05-03 at 16 08 45

PR #27, opened 2017-05-04
Special treatment for SubArray of InvWarpedView

Evizero:
continuation of #26


timholy:
Sorry to be slow here, very nice work!

Evizero:
Uh, this is still untested. Sorry I didn't flag this as WIP. It does work however, so no harm done. I'll make a PR with tests soonish

timholy:
Oh, sorry I failed to notice that!

PR #32, opened 2017-05-17
Make tests pass even when julia is started without --color=yes

tkelman:
kinda messy, but seems to work - Base could use a better scoped API for this at some point


Evizero:
That is an interesting way of doing it. I came up with a much lazier solution here https://github.com/Evizero/Augmentor.jl/blob/master/test/tst_augment.jl#L12-L17  but I think I prefer yours

tkelman:
pkgeval will notice regressions sooner with this version than the skipping tests option

Evizero:
makes sense to me. Thanks for doing this. Will merge once CI passed

PR #34, opened 2017-07-13
Fix reparamterization of CartesianRange on 0.7

timholy:
Doesn't pass tests on 0.7, but that wasn't really my goal here. Fixes https://github.com/JuliaImages/Images.jl/issues/653.


timholy:
(Well, there's a later failure in ImageFiltering due to https://github.com/JuliaMath/FFTW.jl/issues/13, so this won't really get Images even building let alone passing.)

timholy:
Or we could wait for https://github.com/JuliaLang/Compat.jl/pull/377

PR #35, opened 2017-08-16
Fix `restrict` for Lab and other color types

timholy:
Fixes https://github.com/JuliaImages/Images.jl/issues/652


timholy:
Interesting, 0.5 failed to infer the output type. To prevent a possible breakage on 0.5, I just bumped the minimum julia version.

Issue #43, opened 2017-11-13
Update code to reflect upstream changes to Rotations


Evizero:
It looks like the tests fail for new some reason. (I have yet to look into it)

```
autorange: Error During Test
  Got an exception of type MethodError outside of a @test
  MethodError: Rotations.RotMatrix{2,Float64,4}(::StaticArrays.SArray{Tuple{2,2},Float64,2,4}) is ambiguous. Candidates:
    (::Type{Rotations.RotMatrix{N,T,L}})(x::AbstractArray) where {N, T, L} in Rotations at /home/csto/.julia/v0.6/Rotations/src/core_types.jl:78
    (::Type{SA})(a::StaticArrays.StaticArray) where SA<:StaticArrays.StaticArray in StaticArrays at /home/csto/.julia/v0.6/StaticArrays/src/convert.jl:4
  Possible fix, define
    (::Type{Rotations.RotMatrix{N,T,L}})(::StaticArrays.StaticArray)
  Stacktrace:
   [1] macro expansion at /home/csto/.julia/v0.6/ImageTransformations/test/autorange.jl:108 [inlined]
```

Evizero:
solved itself with the merge and tag of https://github.com/FugroRoames/Rotations.jl/pull/50

PR #44, opened 2017-11-14
make sure extrapolation preserves eltype

Evizero:
Something in `Interpolations.jl` must have changed, so this conversion didn't happen upstream anymore

### DIFF
--- a/src/interpolations.jl
+++ b/src/interpolations.jl
@@ -12,10 +12,13 @@ const FloatColorant{T<:AbstractFloat} = Colorant{T}
 @inline _default_fill(::Type{T}) where {T<:FloatColorant} = _nan(T)
 @inline _default_fill(::Type{T}) where {T} = zero(T)
 
+@inline _make_compatible(T, fill) = fill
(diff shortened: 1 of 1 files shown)


timholy:
Most likely https://github.com/JuliaMath/Interpolations.jl/pull/179, a fix for the problems noted in https://github.com/JuliaMath/Interpolations.jl/pull/177.

Evizero:
CI errors are because of #43 (StaticArrays)

timholy:
This will run into problems if you use `NaN` filling with `NoInterp` interpolation, but otherwise it seems safe.

Evizero:
Mhm, I guess it is not too far fetched that one could have a `64x64x1000` array of `1000` independent images that should be transformed the same way. In such a case the third axis should probably be `NoInterp`

That said, if I understand correctly then the `NaN` issue was present before your fix, right? In which case I'd say that this PR is really just about preserving current behaviour. Let's tackle the `NaN` issue when it comes up.

PR #56, opened 2018-08-27
minor cleanups for Julia 1.0

rdeits:
* Use the default Travis script, which should be a bit more future-proof
* Remove some unnecessary blank lines in the readme (this causes the badges to all display in a line instead of in separate paragraphs)
* Bump the ReferenceTests requirement to get some Julia 1.0 fixes


Evizero:
thanks!

PR #58, opened 2018-09-03
 Add image rotation

johnnychen94:
Add an `imrotate` function as a convenient way to replace `warp` in image rotation.

For numerical accuracy, Check: https://github.com/FugroRoames/Rotations.jl/pull/70#issuecomment-418102087


src/warp.jl
```diff
@@ -100,3 +100,46 @@ function warp(img::AbstractArray, tform, args...)
...
+"""
+function imrotate(img::AbstractArray{T}, angle::Real; fill = NaN, method::String = "nearest", bbox::String = "loose")::AbstractArray{T} where T
+
+    # FIXME: Bicubic method does not support N0f8 type.
+    args = []
+    push!(args,
+        if method == "nearest"
+            Constant()
+        elseif method == "bilinear"
+            Linear()
```
Evizero: I am skeptical of this. Why do we need this string glue code when we have the types. In general I think we should avoid string option parameter unless we really need them. I know @timholy has chosen to strings for the padding in `imfilter`. Maybe he has a different opinion on this?

johnnychen94: In my opinion, the reason is simple.

It is kind of inconvenient writing a two line `warp` code for a simple `imrotate` job, not to mention one has to explicitly `using CoordinateTransformations, Interpolations` for this work.

Using String parameters is my personal old-habit from MATLAB, if you have a better idea of implementing this without String keywords, I'm very happy to change it; actually, I find it not concise in parsing these parameters.

RalphAS: See the pattern used for `degree` in `src/interpolations.jl` (in this package) for a good alternative here.
Don't forget to add `imrotate` to the exports list in `src/Interpolations.jl` and drop the reference to a deleted file in `test/runtests.jl`.

johnnychen94:
@Evizero What about this one? `imrotate(img, θ, [indices], [degree = Linear()], [fill = NaN])` as a simple wrap of `warp`.

test/warp.jl
```diff
@@ -387,4 +387,40 @@ ref_img_pyramid_grid = Float64[
...
+
+    @testset "imrotate" begin
+        # imrotate is a wrapper, hence only need to guarantee the interface works as expected
+
+        test_types = (Float32, Float64, N0f8, N0f16)
+        graybar = repeat(range(0,stop=1,length=100),1,100)
+        @testset "interface" begin
+            for T in test_types
+                img = Gray{T}.(graybar)
+                @test_nowarn imrotate(img, 45)
```
Evizero: why `45` if its in radians? What are you testing here?

test/warp.jl
```diff
@@ -387,4 +387,40 @@ ref_img_pyramid_grid = Float64[
...
+                @test_nowarn imrotate(img, 45, Constant())
+                @test_nowarn imrotate(img, 45, Linear())
+                @test_nowarn imrotate(img, 45, axes(img))
+                @test_nowarn imrotate(img, 45, axes(img), Constant())
```
Evizero: since you mention it in the signature of the docs, it would be useful to assert that

```julia
@test imrotate(img, pi) ≈ imrotate(img, pi, Linear(), NaN)
```

test/warp.jl
```diff
@@ -387,4 +387,40 @@ ref_img_pyramid_grid = Float64[
...
+                    p = 2pi*randn()
+                    @test imrotate(img,p) == imrotate(img,p+2pi)
+                end
+            end
+            # FIXME: Floats are not accurate enough due to RotMatrix is not accurate
+            for T in test_types[1:2]
+                img = Gray{T}.(graybar)
+                for i in 1:repeats
+                    p = 2pi*randn()
+                    @test_broken imrotate(img,p) ≈ imrotate(img,p)
```
Evizero: how can this test test possibly be broken? Is this function non-deterministic?

test/warp.jl
johnnychen94: It's a mistake here, in the first version of `imrotate` I didn't use radians but degrees. I'll change this.

test/warp.jl
johnnychen94: Sorry for the wrong comment. It's the behavior of `NaN` since the default filling value is `NaN` is supported. `Float` supports `NaN` and `N0f8` doesn't.
Change to `isequal` shall be fine.

```julia
julia> NaN == NaN
false
julia> isequal(NaN, NaN)
true
```

Evizero: oh, right. good point.

johnnychen94:
Some test fails at present, mainly due to the following fact:

```julia
julia> θ = 2.6927937030769655
julia> mod2pi(θ+2π) == θ
false
```

For this reason,  we can not even require `imrotate(img, θ+ 2pi) ≈ imrotate(img, θ)` holds

I'd like to delete the whole `@testset "numerical"` since `imrotate` is only a wrapper of `warp` and `RotMatrix`. ( To be honest, I don't have a solution for this, even if I have, I don't have time. Perhaps we need to create a more accurate type for `Radius`)

test/warp.jl
johnnychen94: By adding this test, I met one weird problem. Not even know how could this happen.

```julia
# For Float type only
img = Gray{Float32}.(ones(5,5))
a1 = imrotate(img, π/4)
a2 = imrotate(img, π/4, Linear())
isequal(a1, a2) # false
isequal(channelview(a1), channelview(a2)) # true
```

Hence if I want to let the test pass, I need to write
```julia
@test isequal(channelview(imrotate(img, pi)), channelview(imrotate(img, pi, Linear(), NaN)))
```

Evizero: that is strange. maybe leave a todo comment explaining this for some future investigation

src/warp.jl
```diff
@@ -100,3 +100,32 @@ function warp(img::AbstractArray, tform, args...)
...
+julia> imrotate(img, π/4, axes(img))
+
+# rotate with nearest interpolation but without cropping
+julia> imrotate(img, π/4, Constant())
+```
+
+See also [`warp`](@ref).
+"""
+function imrotate(img::AbstractArray{T}, θ::Real, args...) where T
+    θ = floor(mod(θ,2pi)*typemax(Int16))/typemax(Int16) # periodic discretezation
```
johnnychen94: I discretize the radius θ in this way so that `imrotate(img, θ) == imrotate(img, θ+2pi)`. This inaccuracy shall not be the problem in image processing.

But actually I don't know whether this is a good step.

johnnychen94: I want to keep this equality because at some time people will find it surprising that

```julia
img = testimage("cameraman")
imrotate(img,0) .== imrotate(img,2pi) # DimensionMismatch
```

Evizero: fair point

johnnychen94:
Thanks!

Issue #57, opened 2018-08-29
imresize throw StackOverflowError with sz of array type


Evizero:
reported by @johnnychen94 at https://github.com/JuliaImages/Images.jl/issues/742

> The document of `imresize` doesn't say sz must be `Tuple` type, so I just try the MATLAB way of using `imresize`, and it throws an unexpected StackOverflowError.
>
>```julia
>using Images, TestImages
>lena = testimage("lena_gray_256")
>
>imresize(lena, (128,128)) # work as expected
>imresize(lena, [128,128]) # throw StackOverflowError
>```

johnnychen94:
Update:

The document of `imresize` doesn't say sz must be `Tuple{Vararg{Integer}}` type, so I just try the MATLAB way of using `imresize`, and it throws an unexpected StackOverflowError. 
  * use `[128,128]` as size, it's `Array` type.
  * use `ceil.(size(img) .* 0.5)` as size, it is `Tuple{Float64,Float64}` Type

```julia
using Images, TestImages
lena = testimage("lena_gray_256")

imresize(lena, (128, 128)) # work as expected
imresize(lena, [128, 128]) # throw StackOverflowError
imresize(lena, (128.0, 128.0)) # throw StackOverflowError
```

With a temporary solution, just use
```julia
using Images, TestImages
lena = testimage("lena_gray_256")

imresize(lena, (128, 128)) # work as expected
imresize(lena, ceil.(Int, size(lena) .* 0.5)) # work as expected
```

Also,
can we add an `imresize(img, ratio)` API support to this function just like MATLAB? Since Images are always at least 2D, there will be no ambiguity. I think it might be fairly easy to implement.
```julia
imresize(original::AbstractArray, dim1, dimN...) = imresize(original, (dim1,dimN...)) 
imresize(original::AbstractArray, ratio) = imresize(original, round.(Int, size(original).* ratio) # add this line of code
```

Evizero:
Thanks for the thorough report. I agree the current situation is suboptimal. I also like the ratio idea

timholy:
:+1: , thanks indeed @johnnychen94.

But we should use a keyword argument for `ratio`. Otherwise it has ambiguous intent for one-dimensional "images." (JuliaImages has a much firmer commitment to providing good support for images of arbitrary dimensionality than does Matlab.) Or we could use `Percent` but I think a keyword argument would be better.

PR #59, opened 2018-09-04
Let `imresize` throw `MethodError` instead of `StackoverflowError`

johnnychen94:
1. fix issue #57 
2. add a `imresize(img, ratio)` function
3. ~add support for array, `imresize(img, [256,256])` falls back to `imresize(img,(256,256))`.~
3. rewrite the testset of imresize

FIXME:
* `imresize` is not type stable with `UnitRange`, e.g., `@inffered imresize(img, 1:256)` would raise error. I set these test as `@test_broken`

### DIFF
--- a/src/resizing.jl
+++ b/src/resizing.jl
@@ -175,7 +175,9 @@ function imresize(original::AbstractArray; ratio::Real)
     imresize(original, new_size)
 end
 
+# temporary patch for issue #57
 imresize(original::AbstractArray, short_size::Tuple{T,Vararg{T}}) where T<:AbstractFloat = throw(MethodError(imresize, original, short_size))
+
 function imresize(original::AbstractArray, short_size::Tuple)
     len_short = length(short_size)
     len_short > ndims(original) && throw(DimensionMismatch("$short_size has too many dimensions for a $(ndims(original))-dimensional array"))
--- a/src/resizing.jl
+++ b/src/resizing.jl
@@ -175,7 +175,7 @@ function imresize(original::AbstractArray; ratio::Real)
     imresize(original, new_size)
 end
 
-imresize(original::AbstractArray, short_size::Tuple{T,Vararg{T}}) where T<:Real = throw(MethodError(imresize, original, short_size))
+imresize(original::AbstractArray, short_size::Tuple{T,Vararg{T}}) where T<:AbstractFloat = throw(MethodError(imresize, original, short_size))
 function imresize(original::AbstractArray, short_size::Tuple)
     len_short = length(short_size)
     len_short > ndims(original) && throw(DimensionMismatch("$short_size has too many dimensions for a $(ndims(original))-dimensional array"))
--- a/src/resizing.jl
+++ b/src/resizing.jl
@@ -175,10 +175,7 @@ function imresize(original::AbstractArray; ratio::Real)
     imresize(original, new_size)
 end
 
-# temporary patch for issue #57
-imresize(original::AbstractArray, short_size::Tuple{T,Vararg{T}}) where T<:AbstractFloat = throw(MethodError(imresize, original, short_size))
-
-function imresize(original::AbstractArray, short_size::Tuple)
+function imresize(original::AbstractArray, short_size::Union{Indices{M},Dims{M}}) where M
     len_short = length(short_size)
     len_short > ndims(original) && throw(DimensionMismatch("$short_size has too many dimensions for a $(ndims(original))-dimensional array"))
     new_size = ntuple(i -> (i > len_short ? odims(original, i, short_size) : short_size[i]), ndims(original))
--- a/src/resizing.jl
+++ b/src/resizing.jl
@@ -168,8 +168,16 @@ function restrict_indices(r::UnitRange)
 end
 
 # imresize
-imresize(original::AbstractArray, dim1, dimN...) = imresize(original, (dim1,dimN...))
+imresize(original::AbstractArray, dim1::T, dimN::T...) where T<:Union{Integer,AbstractUnitRange} = imresize(original, (dim1,dimN...))
+imresize(original::AbstractArray, short_size::AbstractArray{T,1}) where T <:Union{Integer,AbstractUnitRange} = imresize(original, (short_size...))
 
+function imresize(original::AbstractArray, ratio::Real)
+    # use ceil to avoid 0
+    new_size = ratio > 0 ? ceil.(Int, size(original).* ratio) : throw(ArgumentError("$ratio should be positive"))
+    imresize(original, new_size)
+end
+
+imresize(original::AbstractArray, short_size::Tuple{T,Vararg{T}}) where T<:AbstractFloat = throw(MethodError(imresize, original, short_size))
 function imresize(original::AbstractArray, short_size::Tuple)
     len_short = length(short_size)
     len_short > ndims(original) && throw(DimensionMismatch("$short_size has too many dimensions for a $(ndims(original))-dimensional array"))
@@ -183,15 +191,30 @@ odims(original, i, short_size) = oftype(first(short_size), axes(original, i))
 """
     imresize(img, sz) -> imgr
     imresize(img, inds) -> imgr
+    imresize(img, ratio::Real) -> imgr
 
-Change `img` to be of size `sz` (or to have indices `inds`). This
-interpolates the values at sub-pixel locations. If you are shrinking
-the image, you risk aliasing unless you low-pass filter `img`
-first. For example:
+Change `img` to be of size `sz` (or to have indices `inds`). If `ratio` is used, then
+`sz = ceil(Int, size(img).*ratio)`. This interpolates the values at sub-pixel locations.
+If you are shrinking the image, you risk aliasing unless you low-pass filter `img` first.
 
-    σ = map((o,n)->0.75*o/n, size(img), sz)
-    kern = KernelFactors.gaussian(σ)   # from ImageFiltering
-    imgr = imresize(imfilter(img, kern, NA()), sz)
+# Example:
+```julia
+julia> img = testimage("lena_gray_256") # 256*256
+julia> imresize(img, 128, 128) # 128*128
+julia> imresize(img, 1:128, 1:128) # 128*128
+julia> imresize(img, [128, 128]) # 128*128
+julia> imresize(img, [1:128, 1:128]) # 128*128
+julia> imresize(img, (128, 128)) # 128*128
+julia> imresize(img, (1:128, 1:128)) # 128*128
+julia> imresize(img, (1:128, )) # 128*256
+julia> imresize(img, 128) # 128*256
+julia> imresize(img, 0.5) # 128*128
+```
+
+!!! info
+
+    for N-dimensional image/signal, `imresize(img, arg)` treats `arg::Real` as `ratio`, and `arg::Integer` as `sz`. In the latter case, only the first dimension will be resized.
+
 
 See also [`restrict`](@ref).
 """
--- a/src/resizing.jl
+++ b/src/resizing.jl
@@ -175,8 +175,7 @@ function imresize(original::AbstractArray; ratio::Real)
     imresize(original, new_size)
 end
 
-imresize(original::AbstractArray, short_size::Tuple{T,Vararg{T}}) where T<:AbstractFloat = throw(MethodError(imresize, original, short_size))
-function imresize(original::AbstractArray, short_size::Tuple)
+function imresize(original::AbstractArray, short_size::Tuple{T,Vararg{T}}) where T <: Union{Integer, AbstractUnitRange}
     len_short = length(short_size)
     len_short > ndims(original) && throw(DimensionMismatch("$short_size has too many dimensions for a $(ndims(original))-dimensional array"))
     new_size = ntuple(i -> (i > len_short ? odims(original, i, short_size) : short_size[i]), ndims(original))
--- a/src/resizing.jl
+++ b/src/resizing.jl
@@ -175,7 +175,8 @@ function imresize(original::AbstractArray; ratio::Real)
     imresize(original, new_size)
 end
 
-function imresize(original::AbstractArray, short_size::Tuple{T,Vararg{T}}) where T <: Union{Integer, AbstractUnitRange}
+imresize(original::AbstractArray, short_size::Tuple{T,Vararg{T}}) where T<:Real = throw(MethodError(imresize, original, short_size))
+function imresize(original::AbstractArray, short_size::Tuple)
     len_short = length(short_size)
     len_short > ndims(original) && throw(DimensionMismatch("$short_size has too many dimensions for a $(ndims(original))-dimensional array"))
     new_size = ntuple(i -> (i > len_short ? odims(original, i, short_size) : short_size[i]), ndims(original))
--- a/src/resizing.jl
+++ b/src/resizing.jl
@@ -169,11 +169,9 @@ end
 
 # imresize
 imresize(original::AbstractArray, dim1::T, dimN::T...) where T<:Union{Integer,AbstractUnitRange} = imresize(original, (dim1,dimN...))
-imresize(original::AbstractArray, short_size::AbstractArray{T,1}) where T <:Union{Integer,AbstractUnitRange} = imresize(original, (short_size...))
-
-function imresize(original::AbstractArray, ratio::Real)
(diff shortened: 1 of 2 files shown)


src/resizing.jl
```diff
@@ -182,16 +191,43 @@ odims(original, i, short_size) = oftype(first(short_size), axes(original, i))
...
+- for N-dimensional image/signal, `imresize(img, arg)` treats `arg::Real` as `ratio`, and `arg::Integer` as `sz`. In the latter case, only the first dimension will be resized.
 
-    σ = map((o,n)->0.75*o/n, size(img), sz)
-    kern = KernelFactors.gaussian(σ)   # from ImageFiltering
-    imgr = imresize(imfilter(img, kern, NA()), sz)
+```julia
+julia> img_1d = 0:0.1:1 .|> n0f8
+julia> imresize(img_1d, 3) # 3*10-element Array{N0f8,1}
+julia> imresize(img_1d, 3.0) # 30*30-element Array{N0f8,1}
+```
```
johnnychen94: I think doing this would not introduce ambiguity. For `Integer` it means size, and for other `Real` type it means ratio.

Evizero:
Thanks! Unfortunately it might take me a couple days before I can review this

johnnychen94:
It’s fine, cause this’s not some urgent issue, and I’m going to be very busy this month. Since I’m new to Julia, I’m looking forward to your feedbacks.

src/resizing.jl
```diff
@@ -168,8 +168,17 @@ function restrict_indices(r::UnitRange)
 end
 
 # imresize
-imresize(original::AbstractArray, dim1, dimN...) = imresize(original, (dim1,dimN...))
+imresize(original::AbstractArray, dim1::T, dimN::T...) where T<:Union{Integer,UnitRange} = imresize(original, (dim1,dimN...))
```
Evizero: I think `UnitRange` should be `AbstractUnitRange` here

src/resizing.jl
```diff
@@ -168,8 +168,17 @@ function restrict_indices(r::UnitRange)
...
+imresize(original::AbstractArray, short_size::AbstractArray{T,1}) where T <:Union{Integer,UnitRange} = imresize(original, (short_size...))
```
Evizero: also `AbstractUnitRange`

src/resizing.jl
```diff
@@ -168,8 +168,17 @@ function restrict_indices(r::UnitRange)
...
 
+function imresize(original::AbstractArray, ratio::Real)
+    # use ceil to avoid 0
+    new_size = ratio > 0 ? ceil.(Int, size(original).* ratio) : throw(ArgumentError("$ratio should be positive"))
+    imresize(original, new_size)
+end
+imresize(original::AbstractArray, dim1::Integer) = imresize(original, (dim1,))
```
Evizero: is this method really needed? the vararg one at 171 should cover this case

```julia
julia> bar(x::T, y::T...) where {T} = (x,y...)
bar (generic function with 1 method)

julia> bar(2)
(2,)
```

src/resizing.jl
```diff
@@ -182,16 +191,43 @@ odims(original, i, short_size) = oftype(first(short_size), axes(original, i))
...
     imresize(img, sz) -> imgr
-    imresize(img, inds) -> imgr
+    imresize(img, ratio::Real) -> imgr
+    imresize(img, inds::Integer...) -> imgr
+
+Change `img` to be of size `sz` (or to have indices `inds`). If `ratio` is used, then
+`sz = ceil(Int, size(img).*ratio)`. This interpolates the values at sub-pixel locations.
+If you are shrinking the image, you risk aliasing unless you low-pass filter `img` first.
+
+# Example:
```
Evizero: rather than having two `# Example:` sections, we should just have one `# Examples` section. (see https://docs.julialang.org/en/v1/manual/documentation/)

src/resizing.jl
Evizero: the info block is not indented properly. The body of the block should all be indented 4 spaces to be aligned with the ford `info` in `!!! info`. see https://docs.julialang.org/en/v1/manual/documentation/#Admonitions-1

test/resizing.jl
```diff
@@ -51,68 +51,90 @@
...
-        @test imresize(A, (1, 1)) ≈ reshape([0.25], 1, 1)
-    end
-    A = [1 0; 0 0]
-    @test imresize(A, (2, 1)) ≈ reshape([0.5, 0], (2, 1))
-    @test imresize(A, (1, 2)) ≈ reshape([0.5, 0], (1, 2))
-    A = [1 0 0; 0 0 0; 0 0 0]
-    @test imresize(A, (2, 2)) ≈ [0.75^2 0; 0 0]
+    @testset "Interface" begin
+        function test_imresize_interface(img, outsz, args...)
+            img2 = @test_broken @inferred imresize(img, args...) # FIXME: @inferred failed
```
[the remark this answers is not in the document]
Evizero: Right. I bet its this branch here:

https://github.com/JuliaImages/ImageTransformations.jl/blob/88d4730e59e4b989b2d2f1c99286a099c5936b9f/src/resizing.jl#L220-L224

src/resizing.jl
```diff
@@ -182,16 +191,43 @@ odims(original, i, short_size) = oftype(first(short_size), axes(original, i))
 
 """
     imresize(img, sz) -> imgr
-    imresize(img, inds) -> imgr
+    imresize(img, ratio::Real) -> imgr
+    imresize(img, inds::Integer...) -> imgr
```
Evizero: given how the names are used in the description we should at least switch `inds` with `sz` here. (because `inds` means that the elements are ranges, while `sz` implies the elements are `Integer`). Technically both arguments can be a tuple or a var-arg, but i get the idea here to contrast the `Integer` to the `Real`

test/resizing.jl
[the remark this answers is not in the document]
Evizero: I think we can leave this specific issue for some other time/pr

johnnychen94:
I've made changes according to your feedback. 
1. changed `UnitRange` to `AbstractUnitRange` and deleted unnecessarily codes
2. Recheck the document format. The previously given example is deleted for clarity.

src/resizing.jl
```diff
@@ -168,8 +168,16 @@ function restrict_indices(r::UnitRange)
 end
 
 # imresize
-imresize(original::AbstractArray, dim1, dimN...) = imresize(original, (dim1,dimN...))
+imresize(original::AbstractArray, dim1::T, dimN::T...) where T<:Union{Integer,AbstractUnitRange} = imresize(original, (dim1,dimN...))
+imresize(original::AbstractArray, short_size::AbstractArray{T,1}) where T <:Union{Integer,AbstractUnitRange} = imresize(original, (short_size...))
```
johnnychen94: I'm starting to think, do we really need to support `Array` type size, since size in Julia is  conventionally `Tuple` type

Evizero:
oh wow, seems I forgot about this. sorry. I made a note on my whiteboard and will see that i get to this as soon as i find some time

Evizero:
is the broken type inference a new problem?

johnnychen94:
> is the broken type inference a new problem?

This problem already exists there before I make these changes. I just add these tests here as a reminder.

src/resizing.jl
```diff
@@ -183,15 +191,30 @@ odims(original, i, short_size) = oftype(first(short_size), axes(original, i))
 """
     imresize(img, sz) -> imgr
     imresize(img, inds) -> imgr
+    imresize(img, ratio::Real) -> imgr
+
+Change `img` to be of size `sz` (or to have indices `inds`). If `ratio` is used, then
+`sz = ceil(Int, size(img).*ratio)`. This interpolates the values at sub-pixel locations.
+If you are shrinking the image, you risk aliasing unless you low-pass filter `img` first.
+
+# Example:
```
Evizero:
```suggestion
# Examples
```

src/resizing.jl
```diff
@@ -168,8 +168,16 @@ function restrict_indices(r::UnitRange)
 end
 
 # imresize
-imresize(original::AbstractArray, dim1, dimN...) = imresize(original, (dim1,dimN...))
+imresize(original::AbstractArray, dim1::T, dimN::T...) where T<:Union{Integer,AbstractUnitRange} = imresize(original, (dim1,dimN...))
+imresize(original::AbstractArray, short_size::AbstractArray{T,1}) where T <:Union{Integer,AbstractUnitRange} = imresize(original, (short_size...))
 
+function imresize(original::AbstractArray, ratio::Real)
```
timholy: Needs to be a keyword argument, see https://github.com/JuliaImages/ImageTransformations.jl/issues/57#issuecomment-417214869.

Specifically, what if `ratio = 2`? That's not an `AbstractFloat` (so your `throw(MethodError(...))` below won't catch it) and yet my intention was to get a 2x change in size.

src/resizing.jl
timholy: I don't think we need to support Array, a `MethodError` seems sufficient.

Evizero: i would argue no. there is little upside of supporting a suboptimal interface if it has no upsides.

Once thing that to me seems pretty useful is to support it marked as deprecation (but permanently). This would enable newcomers that are used to a different interface to quickly adapt

johnnychen94:
@Evizero I want to remove the `Array` type size, i.e., still let `imresize(img, [128, 128])` throw MethodError. 

At the time I create this PR I was still thinking in MATLAB way( where `[128,128]` is the most common use case), but now I find supporting this in Julia will damage the overall consistency. (Some who find `imresize(img, [128,128])` worked will try `reshape(ones(5,5),[25,1])`)

src/resizing.jl
```diff
@@ -168,8 +168,16 @@ function restrict_indices(r::UnitRange)
...
+    # use ceil to avoid 0
+    new_size = ratio > 0 ? ceil.(Int, size(original).* ratio) : throw(ArgumentError("$ratio should be positive"))
```
timholy: I tend to have a slight preference for
```julia
ratio > 0 || throw(ArgumentError("$ratio should be positive"))
new_size = ceil.(Int, size(original).* ratio)
```
rather than the trinary operator in such cases, but this is purely stylistic and not necessary to change.

Evizero:
alternatively we could compromise to throw a better warning using a deprecation.

e.g.

```julia
julia> foo(x::Tuple) = x.^2
foo (generic function with 1 method)

julia> @deprecate bar(x::Array) foo((x...,))
bar (generic function with 1 method)

julia> bar([1,2,3])
┌ Warning: `bar(x::Array)` is deprecated, use `foo((x...,))` instead.
│   caller = top-level scope at none:0
└ @ Core none:0
(1, 4, 9)
```

src/resizing.jl
johnnychen94: This is why I put some note on the function document. But yes this would implicitly introduce the possibility of a new bug. (although not so hard to find)
I'll change this.

src/resizing.jl
```diff
@@ -168,8 +168,14 @@ function restrict_indices(r::UnitRange)
...
 # imresize
-imresize(original::AbstractArray, dim1, dimN...) = imresize(original, (dim1,dimN...))
+imresize(original::AbstractArray, dim1::T, dimN::T...) where T<:Union{Integer,AbstractUnitRange} = imresize(original, (dim1,dimN...))
+function imresize(original::AbstractArray; ratio::Real)
+    ratio > 0 || throw(ArgumentError("ratio $ratio should be positive"))
+    new_size = ceil.(Int, size(original) .* ratio) # use ceil to avoid 0
+    imresize(original, new_size)
+end
 
+imresize(original::AbstractArray, short_size::Tuple{T,Vararg{T}}) where T<:AbstractFloat = throw(MethodError(imresize, original, short_size))
```
johnnychen94: Previously `imresize(ones(5,5), 5.0,5.0)` throws `StackOverflowError` instead of `MethodError`. I add this line to capture this.

However, this actually doesn't capture all possibilities, e.g., `imresize(ones(5,5),(5, 5.0))`, but I can't find a simple way to make this happen. So I set `@test_broken imresize(ones(5,5), (5,5.0))`

johnnychen94:
Thanks for your patient reviews and explanations @timholy @Evizero These help me a lot in understanding Julia.

It's quite long since I create this PR, and some chaos happened when I made changes according to your comments.

As a summary, there're still some remaining work with this PR: (already exist, not what this PR introduces)

* `imresize` is not type stable with UnitRange, e.g., `@inffered imresize(img, 1:256)` would raise error. I set these test as `@test_broken`
* issue #57 is only partially fixed: `imresize(img, 256, 256.0)` still throws a `StackOverflowError`. Also set as `@test_broken`

I have to say sorry that I need to focus on my own research projects recently and don't have any more time to fix these two issues (not so urgent I think), so perhaps left these two for new PRs.

johnnychen94:
The codes actually pass the test; I only add one line of comment in the latest commit. Not very clear about the test failure information.

src/resizing.jl
Evizero: i don't get how with your new PR `imresize(ones(5,5), 5.0,5.0)` could possibly throw a stack overflow (if you omit this "temporary patch"). which method would be invoked? I suspect the new line 171 fixed that issue silently by restricting `T`.

concerning `imresize(ones(5,5),(5, 5.0))` you could try adding the same `T` restriction to line 181

johnnychen94: Sorry I made a mistake in expressing this; it's `imresize(ones(5,5), (5.0,5.0))` instead of `imresize(ones(5,5), 5.0,5.0)`

---
As for why I add this not-so-good temporary patch...

1. By changing line 171 `imresize(original::AbstractArray, dim1, dimN...)` to
```julia
imresize(original::AbstractArray, dim1::T, dimN::T...) where T<:Union{AbstractUnitRange, Integer}
```

`imresize(ones(5,5), 5.0, 5.0)` will then throws `MethodError` than `StackOverflowError`.

2. Immediately, one might want to change line 179 `imresize(original::AbstractArray, short_size::Tuple)` to 
```julia
imresize(original::AbstractArray, short_size::Tuple{Vararg{T}}) where T<: Integer
```
to let `imresize(ones(5,5), (5.0,5.0))` throws `MethodError` instead of `StackOverflowError`. (Note here we can't let `T<:Union{Integer, AbstractRange}` here because this will introduce ambiguity with https://github.com/JuliaImages/ImageTransformations.jl/blob/110d3ffab559aee91ebc123c849ca2d768ff14de/src/resizing.jl#L232
However, this will make `imresize(ones(5,5), 1:5)` and `imresize(ones(5,5,5), 1:5, 1:5)` raise a `MethodError`...  
(You  might understand how I'm getting myself crazy and make it a chaos here now...)

Evizero: what about 

```julia
imresize(original::AbstractArray{T,N}, short_size::Union{Indices{M},Dims{M}}) where {T,N,M}
```

johnnychen94:
Thanks for pointing out the solution! @Evizero I wasn't know this before you told me; playing with type multiple dispatch is still a little tricky to me. I think I need to reread the documents more carefully at some time.

src/resizing.jl
```diff
@@ -183,15 +188,23 @@ odims(original, i, short_size) = oftype(first(short_size), axes(original, i))
...
-interpolates the values at sub-pixel locations. If you are shrinking
-the image, you risk aliasing unless you low-pass filter `img`
-first. For example:
+Change `img` to be of size `sz` (or to have indices `inds`). If `ratio` is used, then
+`sz = ceil(Int, size(img).*ratio)`. This interpolates the values at sub-pixel locations.
+If you are shrinking the image, you risk aliasing unless you low-pass filter `img` first.
 
-    σ = map((o,n)->0.75*o/n, size(img), sz)
-    kern = KernelFactors.gaussian(σ)   # from ImageFiltering
-    imgr = imresize(imfilter(img, kern, NA()), sz)
```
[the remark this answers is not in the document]
johnnychen94: It's a good example, but I think it's more clear to remove the `imfilter` part and remain the core usage of `imresize`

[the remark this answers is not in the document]
Dsuke-K: I hope the example remains as is since it seems to be the most straightforward example for anti-aliasing.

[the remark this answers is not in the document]
johnnychen94: Ok I'll keep this.

Evizero:
thanks!

PR #61, opened 2018-09-07
fix _getindex type stability

### LINKED ISSUE #60: Serious regression of warp!
Some serious memory allocation and regression in Julia 0.7. Here is a MWE.

```
julia> img = rand(100, 100);
julia> tfm = recenter(RotMatrix(-pi/4), center(img));
```
In Julia 0.6,
```
julia> @btime warp(img, tfm);
  422.876 μs (5 allocations: 157.72 KiB)
```
In Julia 0.7,
```
julia> @btime warp(img, tfm);
  8.478 ms (261626 allocations: 9.07 MiB)
```
Looking at the code of `warp!`,
```
function warp!(out, img::AbstractExtrapolation, tform)
    @inbounds for I in CartesianIndices(axes(out))
        out[I] = _getindex(img, tform(SVector(I.I)))
    end
    out       <==== should not return anything
end
```
Obviously, `warp!` should not return `out`. But eliminating the return only helps marginally. But I cannot dig any further. Might be some problem with `_getindex`.

Evizero:
closes #60 reported by @jianghaizhu

in 1.0:

```
julia> @btime warp(img, tfm);
  344.735 μs (5 allocations: 157.72 KiB)
```

### DIFF
--- a/src/ImageTransformations.jl
+++ b/src/ImageTransformations.jl
@@ -32,8 +32,7 @@ include("warp.jl")
 include("warpedview.jl")
 include("invwarpedview.jl")
 
-#@inline _getindex(A, v::StaticVector) = A[convert(Tuple, v)...]
-@inline _getindex(A, v::StaticVector) = getindex(A, v...)
+@inline _getindex(A, v::StaticVector) = A[Tuple(v)...]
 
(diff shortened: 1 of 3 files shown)


jianghaizhu:
Thanks! That works.

PR #62, opened 2018-09-19
Update for Interpolations 0.9

timholy:
The part of this I'm the least sure about is the Project.toml file. Should we be bumping the version number "pre-release" like this? It's also a bit unfortunate that I have those unbounded `≥ 0.2"` in the compat section (makes me think we want to transition to 1.x releases for packages sooner rather than later).

### DIFF
--- a/REQUIRE
+++ b/REQUIRE
@@ -1,7 +1,7 @@
 julia 0.7
 ImageCore 0.7
 CoordinateTransformations 0.4.0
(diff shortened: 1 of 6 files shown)


timholy:
I removed the Project.toml file because we also need test requirements, and currently I am not aware of a way to make those platform-specific (I'm thinking of ImageMagick vs QuartzImageIO).

PR #65, opened 2019-02-01
Fix tests on Julia 1.1

timholy:
This fixes tests that broke with Julia 1.1.


timholy:
Breakage on nightly was due to https://github.com/JuliaLang/julia/pull/30867; I added a 2nd commit to fix it.

PR #70, opened 2019-09-13
Expand docs for `restrict`

timholy:
Fixes https://github.com/JuliaImages/Images.jl/issues/818. CC @ianshmean, let me know if this seems adequate. As you can see it's not an entirely trivial set of concepts, so now the docs for `restrict` occupy a full screen. However, verbosity is probably the lesser of the two evils.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -22,7 +22,7 @@ CoordinateTransformations = ">= 0.4.0"
 FixedPointNumbers = ">= 0.3"
 ImageCore = ">= 0.7"
 Interpolations = ">= 0.9"
-julia = ">= 1.0"
+julia = "1"
 
 [extras]
 ImageMagick = "6218d12a-5da1-5696-b52f-db25d2ecc6d1"
--- a/src/resizing.jl
+++ b/src/resizing.jl
@@ -1,12 +1,71 @@
 """
     restrict(img[, region]) -> imgr
 
-Reduce the size of `img` by two-fold along the dimensions listed in
-`region`, or all spatial coordinates if `region` is not specified.  It
-anti-aliases the image as it goes, so is better than a naive summation
+Reduce the size of `img` by approximately two-fold along the dimensions listed in
+`region`, or all spatial coordinates if `region` is not specified.
+The term `restrict` is taken from the coarsening operation of algebraic multigrid
+methods; it is the adjoint of "prolongation" (which is essentially interpolation).
+`restrict` anti-aliases the image as it goes, so is better than a naive summation
 over 2x2 blocks.
+The implementation of `restrict` has been tuned for performance, and should be
+a fast method for constructing [pyramids](https://en.wikipedia.org/wiki/Pyramid_(image_processing)).
+
+If `l` is the size of `img` along a particular dimension, `restrict` produces an
+array of size `(l+1)÷2` for odd `l`,
+and `l÷2 + 1` for even `l`. See the example below for an explanation.
 
 See also [`imresize`](@ref).
+
+# Example
+
+```julia
+a_course = [0, 1, 0.3]
+```
+If we were to interpolate this at the halfway points, we'd get
+```julia
+a_fine = [0, 0.5, 1, 0.65, 0.3]
+```
+Note that `a_fine` is obtained from `a_course` via the *prolongation* operator `P` as
+`P*a_course`, where
+```julia
+P = [1   0   0;      # this line "copies over" the first point
+     0.5 0.5 0;      # this line takes the mean of the first and second point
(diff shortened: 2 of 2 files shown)


ianshmean:
This is great! Thanks for elaborating! 
I was still a little slow on understanding why restrict of even dims goes to odd, so I thought I'd make this to help illustrate. I'm sure it could be improved for language and design, if it's considered helpful.

restrict_examples


```julia
using PyPlot

subplots_adjust(hspace=0.7)

subplot(2,1,1)
plot([1,2,3,4], [2,2,2,2], "o", label="Original")
plot([1.25,2.5,3.75], [1,1,1], "o", label="Restricted")
title("Restrict on an even-length dimension")
ylim(0,3);
gca().get_yaxis().set_visible(false)
legend(ncol=2)

subplot(2,1,2)
plot([1,2,3,4,5], [2,2,2,2,2], "o")
plot([1,3,5], [1,1,1], "o")
title("Restrict on an odd-length dimension")
ylim(0,3)
gca().get_yaxis().set_visible(false)
```

timholy:
I'm not certain I fully understand the graphic (sorry). My brief explanation is to start from prolongation (interpolation), because it's clear what to do, and then take the adjoint to figure out how `restrict` is supposed to work.

Let's say with start with the 3-vector `a_coarse`. But this time, instead of taking up to a vector of length 5 (by interpolating at half-grid points), we take it to a vector of length 4: we interpolate at 1.25, 1.75, 2.25, and 2.75.

```julia
julia> a_coarse = [0, 1, 0.3]
3-element Array{Float64,1}:
 0.0
 1.0
 0.3

julia> using Interpolations    # let's do it using traditional interpolation

julia> itp = interpolate(a_coarse, BSpline(Linear()))
3-element interpolate(::Array{Float64,1}, BSpline(Linear())) with element type Float64:
 0.0
 1.0
 0.3

julia> a_fine = [itp(1.25), itp(1.75), itp(2.25), itp(2.75)]   # evenly-spaced points (half a grid apart, but to get an even-length result we do it at quarter-grid points)
4-element Array{Float64,1}:
 0.25 
 0.75 
 0.825
 0.475

julia> P = [0.75 0.25 0;      # do the same using an operator
            0.25 0.75 0;
            0 0.75 0.25;
            0 0.25 0.75]
4×3 Array{Float64,2}:
 0.75  0.25  0.0 
 0.25  0.75  0.0 
 0.0   0.75  0.25
 0.0   0.25  0.75

julia> P*a_coarse        # same as a_fine
4-element Array{Float64,1}:
 0.25 
 0.75 
 0.825
 0.475

julia> P'      # check the adjoint
3×4 LinearAlgebra.Adjoint{Float64,Array{Float64,2}}:
 0.75  0.25  0.0   0.0 
 0.25  0.75  0.75  0.25
 0.0   0.0   0.25  0.75

julia> using ImageTransformations     # now let's see if they are the same

julia> restrict(a_fine)
3-element Array{Float64,1}:
 0.1875            
 0.6812499999999999
 0.28125           

julia> (P'*a_fine)/2
3-element Array{Float64,1}:
 0.1875            
 0.6812499999999999
 0.28125           
```

timholy:
Also worth pointing out:

> restrict of even dims goes to odd

```julia
julia> restrict(1:10)
6-element Array{Float64,1}:
 0.625
 2.5  
 4.5  
 6.5  
 8.5  
 4.875
```
It *can* go to odd but it doesn't have to. The rule really is that under prolongation, `n` goes to either `2n-1` or `2n-2` (your choice). Consequently, under restriction both `2n-1` and `2n-2` must go to `n`.

timholy:
Now I get it! Actually on the even-length one, it's more like the restricted one is at 1, 2, 3, and the original is 1.25, 1.75, 2.25, and 2.75. There might be room for an alternate implementation---it might not cut into the edges? But one has to decide on some boundary conditions for the prolongation.

Anyway, I'll merge this, and more improvements can come later.

ianshmean:
Great. I really appreciate the explanations.

For some context around my need, to help with future functionality ideas.. I'm trying to encode half-sized versions of videos with VideoIO using the H264 codec, which doesn't support odd-length image dimensions. The original images are typical camera image sizes, i.e. 1024x1280, and thus I hoped that restrict would be an efficient way to get half-sized images. (The way I'm doing it for now, is to restrict and ignore the last row/col, given speed is more important than image FOV accuracy for this application)

timholy:
Do you need antialiasing? (i.e., do you care if you get [Moire patterns](https://photographylife.com/what-is-moire)?) If not, summing in 2x2 blocks might be fine.

PR #76, opened 2019-11-23
Add bounds on all dependencies

timholy:
Closes #71
Closes #72
Closes #73
Closes #74

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -17,12 +17,16 @@ StaticArrays = "90137ffa-7385-5640-81b9-e52037218182"
 
 [compat]
 AxisAlgorithms = "1.0"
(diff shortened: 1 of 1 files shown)


timholy:
In the absence of complaints I'll merge this soon.

johnnychen94:
Some version is outdated here, for example, according to https://github.com/JuliaGraphics/ColorVectorSpace.jl/releases only ColorVectorSpace `v0.6`, `v0.7` is compatible with `Julia >= v0.7`. But this isn't a big issue, since it would always throws incompatibility error for `ColorVectorSpace < v0.6" (with the existence of `julia = "1"`).

PR #78, opened 2019-12-14
update Julia version in CI

johnnychen94:
Seems 1.3 is broken here, @timholy can you help to fix that?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,7 +5,7 @@ os:
   - osx
 julia:
   - 1.0
-  - 1.1
+  - 1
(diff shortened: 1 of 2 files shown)


johnnychen94:
The only package difference in Julia 1.3 build is SpecialFunctions `v0.9.0`

timholy:
> Seems 1.3 is broken here, @timholy can you help to fix that?

They were all just annoying printing issues. I've switched to a test format that's a little more robust, at the cost of doing less "real" checking. See #80

PR #79, opened 2019-12-14
specialize imrotate for 0, 90, 180, 270 degrees

johnnychen94:
```julia
img = testimage("cameraman") # (512, 512)

# Before: (514, 514)
# After: (512, 512)
size(imrotate(img, pi/2))

# Before: sizes are not equal
# After: true
imrotate(imrotate(imrotate(imrotate(img, pi/2), pi/2), pi/2), pi/2) == img

# Before: sizes are not equal
# After: true
imrotate(imrotate(img, pi), pi) == img
```


johnnychen94:
I believe this PR is ready for review and merge

Evizero:
is it type stable? was it before this PR?

johnnychen94:
The type instability issue is actually detected in #58 and my current knowledge doesn't support me to fix it

test/warp.jl
```diff
@@ -427,15 +427,26 @@ ref_img_pyramid_grid = Float64[
                 @test_nowarn imrotate(img, π/4, Linear())
                 @test_nowarn imrotate(img, π/4, axes(img))
                 @test_nowarn imrotate(img, π/4, axes(img), Constant())
-                @test isequal(channelview(imrotate(img,π/4)), channelview(imrotate(img, π/4, Linear()))) # TODO: if we remove channelview the test will break for Float
+                @test isequal(channelview(imrotate(img, π/4)), channelview(imrotate(img, π/4, Linear()))) # NaN != NaN
```
timholy: The meaning of this comment is unclear. Do you mean that you wanted to write `==` but instead have to use `isequal`?

timholy (commented):
This looks good except for concerns about non-inferrability. I suppose an alternative would be to write specialized functions (`rotate90`, `rotate180`, `rotate270`) for specific rotations.

johnnychen94:
Thanks for the comment, I'm already aware of the type stability issue, I'll come back and try to fix that later when I finished the upgrading of [Augmentor.jl](https://github.com/Evizero/Augmentor.jl) made by @Evizero (for my own research interest)

kimikage:
BTW, 360 ° seems to be missing.

src/warp.jl
```diff
@@ -124,8 +124,28 @@ julia> imrotate(img, π/4, Constant())
...
-function imrotate(img::AbstractArray{T}, θ::Real, args...) where T
-    θ = floor(mod(θ,2pi)*typemax(Int16))/typemax(Int16) # periodic discretezation
-    tform = recenter(RotMatrix{2}(θ), center(img))
-    warp(img, tform, args...)
+function imrotate(img::AbstractMatrix{T}, θ::Real, args...) where T
+    # 1. discretize periodic for numerical stability to make sure
+    # imrotate(img, θ+2pi) == imrotate(img, θ)
+    # 2. typemax(Int16) is 32767, we choose 32760 to make sure the
+    # discretization result of pi/2 is exactly pi/2 (or 90°)
+    max_num_angles = 32760
```
kimikage: Although the same goes for the current code, why don't you use 32768?

src/warp.jl
```diff
@@ -124,8 +124,28 @@ julia> imrotate(img, π/4, Constant())
...
+    θ = round(Int, 180*floor(mod(θ, 2pi)/pi*max_num_angles)/max_num_angles)
+    tform = recenter(RotMatrix{2}(θ/180*pi), center(img))
```
kimikage: Don't use the θ! :scream:

src/warp.jl
johnnychen94: Well... It's like I randomly choose the first number that meets the requirement, and I feel this choice doesn't matter for image processing tasks.

src/warp.jl
```diff
@@ -124,8 +124,28 @@ julia> imrotate(img, π/4, Constant())
...
+        idx = StepRange.(axes(img))
+        perm_img = PermutedDimsArray(img, (2, 1))
+        return view(perm_img, idx[2], reverse(idx[1]))
+    elseif θ == 180
+        idx = map(i->1:1:length(i), axes(img))
+        return view(img, reverse(idx[1]), reverse(idx[2]))
+    elseif θ == 270
+        idx = StepRange.(axes(img))
+        perm_img = PermutedDimsArray(img, (2, 1))
+        return view(perm_img, reverse(idx[2]), idx[1])
```
mileslucas: any reason to not use some of the tools in base?

https://docs.julialang.org/en/v1/base/arrays/#Base.rot180
https://docs.julialang.org/en/v1/base/arrays/#Base.rotl90
https://docs.julialang.org/en/v1/base/arrays/#Base.rotr90

kimikage: I don't think @johnnychen94 want to reallocate the memory and relocate the elements.:smile:

johnnychen94: True, and there won't be type stability issues if we collect them and return an  `Array`; but that would be slower than views.

I don't have a good fix in mind for this at the moment.

kimikage:
BTW, I think it is better to round the rotation matrix instead of quantizing the angle.

johnnychen94:
That sounds promising to me!

mkitti:
It seems a bit dangerous to specialize on an approximation for an irrational number. Maybe it would be better to have a degree based version of `imrotate` and specialize on multiples of 90 like the title suggests? The discretized `imrotate` could then map to that.

Otherwise, I would really appreciate something like this. It's absence was duly noted when attempting something AoC2020 Day 20: https://adventofcode.com/2020/day/20

mkitti:
`sinpi` and `cospi` could also have some relevance here.
https://docs.julialang.org/en/v1/base/math/#Base.Math.sinpi

timholy:
I'll let @johnnychen94 decide what to do here, but I put all the tests from this PR into #149 and they pass.

PR #85, opened 2019-12-29
Rely on ImageCore for Colors&FixedPointNumbers

timholy:
Implements https://github.com/JuliaImages/Images.jl/issues/802

Closes #81
Closes #82
Closes #83
Closes #84


johnnychen94:
Perhaps we need to modify the General registry as well? Otherwise, there might be the same problem in https://github.com/JuliaGraphics/ColorTypes.jl/pull/143#issuecomment-567150683 -- Another set of maintainance work :(

timholy:
I recently learned about (& plan to use) [RetroCap.jl](https://github.com/bcbi/RetroCap.jl).

PR #89, opened 2020-01-27
specialize tweight on FixedPoint

johnnychen94:
Currently, The return eltype of `ImageTransformations.box_extrapolation(img, Flat())` is `Gray{N0f8}` for `Gray{N0f8}` input but `Float64` for `N0f8`. To keep consistency, I add a specialization method here. (Technically it's a type piracy but I think it's safe to do so here)

C.f. https://github.com/Evizero/Augmentor.jl/issues/39

cc: @Evizero


src/interpolations.jl
```diff
@@ -15,6 +15,7 @@ const FloatColorant{T<:AbstractFloat} = Colorant{T}
 @inline _make_compatible(T, fill) = fill
 @inline _make_compatible(::Type{T}, fill::Number) where {T} = T(fill)
 
+Interpolations.tweight(A::AbstractArray{C}) where C<:FixedPoint = C
```
timholy: I use `C` for `Colorant` and `T` for numbers

johnnychen94: Of course!

johnnychen94:
I then realize that I don't have write permission to this repository when I'm about to merge this PR..

src/interpolations.jl
```diff
@@ -16,6 +16,12 @@ const FloatColorant{T<:AbstractFloat} = Colorant{T}
 @inline _make_compatible(::Type{T}, fill::Number) where {T} = T(fill)
 
 Interpolations.tweight(A::AbstractArray{C}) where C<:Colorant{T} where T = T
+if all(map((FixedPoint, Normed)) do T
+        isempty(methodswith(AbstractArray{<:T}, Interpolations.tweight))
+    end)
+    # Reached when upstream package (e.g., FixedPoint) doesn't provide a specialization
+    Interpolations.tweight(A::AbstractArray{T}) where T<:FixedPoint = T
+end
```
johnnychen94: This is what I can come up with so far to minimize the danger of type piracy (and incremental compilation). Do you have any better ideas? @timholy

timholy: If this is really an issue, I suspect we need to add a `Requires` dependency to Interpolations and put this in a `@require` block there. This solution seems likely to be fragile with respect to load order, and may not be fully correct wrt precompilation.

johnnychen94: > may not be fully correct wrt precompilation.

`Requires` will disable the precompilation of this line of code, and thus won't get any trouble, am I understanding it correctly?

timholy: Correct. I'm also not completely clear on whether the precompiled module would retain this runtime check of the method tables; I think it wouldn't, it would just store the answer at the time when it happened to run. In contrast, by putting it into the `__init__` method it's guaranteed to be correct.

johnnychen94:
This one is possibly no longer needed by Augmentor, so I'm closing it.

PR #93, opened 2020-03-13
Allow multi-dimensional ratio in imresize

mileslucas:
This changes `imresize` to be able to take a ratio with multiple dimensions. 

EG

```julia
imresize(ones(3, 3), ratio=(3, 2))  # -> (9, 6)
```

I've added a line to the docstring showing this and a couple tests (that pass!)

Apologies for some of the superfluous changes from auto-formatting. Please let me know if they should be reverted.

### DIFF
--- a/src/resizing.jl
+++ b/src/resizing.jl
@@ -77,13 +77,13 @@ end
 
 function restrict(A::AbstractArray{T,N}, dim::Integer) where {T,N}
     indsA = axes(A)
-    newinds = ntuple(i->i == dim ? restrict_indices(indsA[dim]) : indsA[i], Val(N))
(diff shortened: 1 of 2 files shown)


mileslucas:
Also, I've made sure it's still type stable via an `@inferred` and `@code_warntype` in my REPL.

johnnychen94:
The test failure is unrelated, so as long as those white-spaces reverted, I think it's good to merge.

mileslucas:
Yeah, I noticed it was auto-formatting one of the files and started using the key one to override it, but I had already committed the first set of changes. 

I’ll revert those changes, soon. Thanks for the quick response!

johnnychen94:
Thanks!

PR #103, opened 2020-12-11
Tolerate diffs in Interpolations.Constant printing

timholy:
Interpolations 0.13.1 introduced parametrizations for `Constant`.


johnnychen94:
Hmmm, interesting... I must have a broken local environment here... I saw more failures than what's fixed here.

timholy:
For me, broken environments are the norm, not the exception :laughing: . "Let's see, which devved package is causing the problem this time?"

johnnychen94 (approved):

PR #112, opened 2021-01-28
skip _round for general functions

### LINKED ISSUE #110: More generic warp
Since the introduction of `_round` (commit a4f219020ad142c5a5bcdb03486e092fd8d9e02b), `warp` (or rather `autorange`) only accepts `<:CoordinateTransformations.Transformation`. Is there any way we can allow for more generic transformations? I, for one, use `Interpolations` to describe more complicated transformations (from camera calibrations), which now do not work.

Of course, one alternative would be to increase the kind of transformations `CoordinateTransformations` can produce.

yakir12:
fixes #110

### DIFF
--- a/src/autorange.jl
+++ b/src/autorange.jl
@@ -91,6 +91,8 @@ function _round(tform::T; kwargs...) where T<:CoordinateTransformations.Transfor
     end
     T(rounded_fields...)
 end
+_round(tform::ComposedFunction; kwargs...) = _round(tform.outer; kwargs...) ∘ _round(tform.inner; kwargs...)
+_round(tform; kwargs...) = tform
 
 __round(x; kwargs...) = x
 __round(x::AbstractArray; digits=_default_digits(eltype(x)), kwargs...) = round.(x; digits=digits, kwargs...)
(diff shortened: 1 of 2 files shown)


timholy:
Requires a test. As the discussion in #110 shows, this was almost much more of a problem of inadequate test coverage than inadequate functionality, so just fixing the functionality is far less important than providing the test.

src/autorange.jl
```diff
@@ -91,6 +91,8 @@ function _round(tform::T; kwargs...) where T<:CoordinateTransformations.Transfor
     end
     T(rounded_fields...)
 end
+_round(tform::ComposedFunction; kwargs...) = _round(tform.outer; kwargs...) ∘ _round(tform.inner; kwargs...)
```
johnnychen94, line 94: `ComposedFunction` exists only in Julia 1.6

```suggestion
if isdefined(Base, :ComposedFunction)
    _round(tform::ComposedFunction; kwargs...) = _round(tform.outer; kwargs...) ∘ _round(tform.inner; kwargs...)
end
```

PR #114, opened 2021-03-10
Fix rounding issue

matsueushi:
The current default number of digits for rounding (used by `autorange`) seems to be too large for rotation matrix calculation. Currently `cos(pi/2)` is rounded up to non-zero.
```julia
julia> round(cos(pi/2); digits=16)
1.0e-16

julia> round(cos(pi/2); digits=15)
0.0
```

I also modified the tests for `autorange` to use trigonometric functions with degree to evaluate expected values because the combination of `ceil` and usual trigonometric function may give unexpected results, for example,
```julia
julia> ϕ, indh, indw = pi/3,-1:10,0:20
(1.0471975511965976, -1:10, 0:20)

julia> cos(ϕ)*last(indh) - sin(ϕ)*first(indw)
5.000000000000001

julia> ceil(cos(ϕ)*last(indh) - sin(ϕ)*first(indw))
6.0
```

Hopefully this also fixes https://github.com/Evizero/Augmentor.jl/issues/77.

### DIFF
--- a/src/autorange.jl
+++ b/src/autorange.jl
@@ -81,8 +81,8 @@ try_static(tfm::Translation{<:AbstractVector}, img::AbstractArray{T,N}) where {T
 # https://github.com/JuliaImages/ImageTransformations.jl/issues/104
 _default_digits(::Type{T}) where T<:Number = _default_digits(floattype(T))
 # these constants come from eps() digits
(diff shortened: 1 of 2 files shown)


johnnychen94:
Nice detective work! Thanks for fixing this.

matsueushi:
Thanks!

PR #115, opened 2021-03-28
Add `method` keyword for `imrotate`/`imresize`/`warp` to allow custom interpolation method

### LINKED ISSUE #111: More Interpolation Options in Imrotate
I was trying to implement more complicated interpolation options in imrotate, but seem to be restricted to passing arguments accepted by the BSpline interpolation class. Is it possible to use other interpolation classes like below?

using Interpolations imrotate(img,1.2,Lanczos4OpenCV())

(this produces an error message)

andrew-saydjari:
Fixes #111 

Added Lanczos4OpenCV support for both imrotate and imresize. Simply created methods which also take a Lanczos4OpenCV argument. Please advise if additions to test or documentation are desired.

### DIFF
--- a/src/resizing.jl
+++ b/src/resizing.jl
@@ -227,18 +227,18 @@ function restrict_indices(r::UnitRange)
 end
 
 # imresize
-imresize(original::AbstractArray, dim1::T, dimN::T...) where T<:Union{Integer,AbstractUnitRange} = imresize(original, (dim1,dimN...))
-function imresize(original::AbstractArray; ratio)
+imresize(original::AbstractArray, dim1::T, dimN::T...; kwargs...) where T<:Union{Integer,AbstractUnitRange} = imresize(original, (dim1,dimN...))
+function imresize(original::AbstractArray; ratio, kwargs...)
     all(ratio .> 0) || throw(ArgumentError("ratio $ratio should be positive"))
     new_size = ceil.(Int, size(original) .* ratio) # use ceil to avoid 0
-    imresize(original, new_size)
+    imresize(original, new_size; kwargs...)
 end
 
-function imresize(original::AbstractArray, short_size::Union{Indices{M},Dims{M}}) where M
+function imresize(original::AbstractArray, short_size::Union{Indices{M},Dims{M}}; kwargs...) where M
     len_short = length(short_size)
     len_short > ndims(original) && throw(DimensionMismatch("$short_size has too many dimensions for a $(ndims(original))-dimensional array"))
     new_size = ntuple(i -> (i > len_short ? odims(original, i, short_size) : short_size[i]), ndims(original))
-    imresize(original, new_size)
+    imresize(original, new_size; kwargs...)
 end
 
 odims(original, i, short_size::Tuple{Integer,Vararg{Integer}}) = size(original, i)
@@ -279,7 +279,7 @@ function imresize(original::AbstractArray{T,0}, new_inds::Tuple{}) where T
     copyto!(similar(original, Tnew), original)
 end
 
-function imresize(original::AbstractArray{T,N}, new_size::Dims{N}) where {T,N}
+function imresize(original::AbstractArray{T,N}, new_size::Dims{N}; kwargs...) where {T,N}
     Tnew = imresize_type(first(original))
     inds = axes(original)
     if map(length, inds) == new_size
@@ -290,31 +290,16 @@ function imresize(original::AbstractArray{T,N}, new_size::Dims{N}) where {T,N}
             copyto!(dest, CartesianIndices(axes(dest)), original, CartesianIndices(inds))
         end
     else
-        imresize!(similar(original, Tnew, new_size), original)
+        imresize!(similar(original, Tnew, new_size), original; kwargs...)
     end
 end
 
-function imresize(original::AbstractArray{T,N}, new_size::Dims{N}, aitp::Interpolations.AbstractLanczos) where {T,N}
-    Tnew = imresize_type(first(original))
-    inds = axes(original)
-    if map(length, inds) == new_size
-        dest = similar(original, Tnew, new_size)
-        if axes(dest) == inds
-            copyto!(dest, original)
-        else
-            copyto!(dest, CartesianIndices(axes(dest)), original, CartesianIndices(inds))
-        end
-    else
-        imresize!(similar(original, Tnew, new_size), original, aitp)
-    end
-end
-
-function imresize(original::AbstractArray{T,N}, new_inds::Indices{N}) where {T,N}
+function imresize(original::AbstractArray{T,N}, new_inds::Indices{N}; kwargs...) where {T,N}
     Tnew = imresize_type(first(original))
     if axes(original) == new_inds
         copyto!(similar(original, Tnew), original)
     else
-        imresize!(similar(original, Tnew, new_inds), original)
+        imresize!(similar(original, Tnew, new_inds), original; kwargs...)
     end
 end
 
@@ -328,15 +313,9 @@ imresize_type(c::Gray) = Gray{imresize_type(gray(c))}
 imresize_type(c::FixedPoint) = typeof(c)
 imresize_type(c) = typeof((c*1)/1)
 
-function imresize!(resized::AbstractArray{T,N}, original::AbstractArray{S,N}) where {T,S,N}
+function imresize!(resized::AbstractArray{T,N}, original::AbstractArray{S,N}; method::Interpolations.InterpolationType=BSpline(Linear()), kwargs...) where {T,S,N}
     # FIXME: avoid allocation for interpolation
-    itp = interpolate(original, BSpline(Linear()))
-    imresize!(resized, itp)
-end
-
-function imresize!(resized::AbstractArray{T,N}, original::AbstractArray{S,N}, aitp::Interpolations.AbstractLanczos) where {T,S,N}
-    # FIXME: avoid allocation for interpolation
-    itp = interpolate(original, aitp)
+    itp = interpolate(original, method)
     imresize!(resized, itp)
 end
 
--- a/src/resizing.jl
+++ b/src/resizing.jl
@@ -263,8 +263,9 @@ julia> imresize(img, (128, 128)) # 128*128
 julia> imresize(img, (1:128, 1:128)) # 128*128
 julia> imresize(img, (1:128, )) # 128*256
 julia> imresize(img, 128) # 128*256
-julia> imresize(img, ratio = 0.5) #
+julia> imresize(img, ratio = 0.5) #128*128
 julia> imresize(img, ratio = (2, 1)) # 256*128
+julia> imresize(img, (128,128), Lanczos4OpenCV()) #128*128
 
 σ = map((o,n)->0.75*o/n, size(img), sz)
 kern = KernelFactors.gaussian(σ)   # from ImageFiltering
--- a/src/resizing.jl
+++ b/src/resizing.jl
@@ -240,6 +240,7 @@ function imresize(original::AbstractArray, short_size::Union{Indices{M},Dims{M}}
     new_size = ntuple(i -> (i > len_short ? odims(original, i, short_size) : short_size[i]), ndims(original))
     imresize(original, new_size)
 end
+
 odims(original, i, short_size::Tuple{Integer,Vararg{Integer}}) = size(original, i)
 odims(original, i, short_size::Tuple{}) = axes(original, i)
 odims(original, i, short_size) = oftype(first(short_size), axes(original, i))
@@ -262,7 +263,7 @@ julia> imresize(img, (128, 128)) # 128*128
 julia> imresize(img, (1:128, 1:128)) # 128*128
 julia> imresize(img, (1:128, )) # 128*256
 julia> imresize(img, 128) # 128*256
-julia> imresize(img, ratio = 0.5) # 
+julia> imresize(img, ratio = 0.5) #
 julia> imresize(img, ratio = (2, 1)) # 256*128
 
 σ = map((o,n)->0.75*o/n, size(img), sz)
@@ -292,6 +293,21 @@ function imresize(original::AbstractArray{T,N}, new_size::Dims{N}) where {T,N}
     end
 end
 
+function imresize(original::AbstractArray{T,N}, new_size::Dims{N}, aitp::Interpolations.Lanczos4OpenCV) where {T,N}
+    Tnew = imresize_type(first(original))
+    inds = axes(original)
+    if map(length, inds) == new_size
+        dest = similar(original, Tnew, new_size)
+        if axes(dest) == inds
+            copyto!(dest, original)
+        else
+            copyto!(dest, CartesianIndices(axes(dest)), original, CartesianIndices(inds))
+        end
+    else
+        imresize!(similar(original, Tnew, new_size), original, aitp)
+    end
+end
+
 function imresize(original::AbstractArray{T,N}, new_inds::Indices{N}) where {T,N}
     Tnew = imresize_type(first(original))
     if axes(original) == new_inds
@@ -314,6 +330,12 @@ imresize_type(c) = typeof((c*1)/1)
 function imresize!(resized::AbstractArray{T,N}, original::AbstractArray{S,N}) where {T,S,N}
     # FIXME: avoid allocation for interpolation
     itp = interpolate(original, BSpline(Linear()))
+    imresize!(resized, itp)
+end
+
+function imresize!(resized::AbstractArray{T,N}, original::AbstractArray{S,N}, aitp::Interpolations.Lanczos4OpenCV) where {T,S,N}
+    # FIXME: avoid allocation for interpolation
+    itp = interpolate(original, aitp)
     imresize!(resized, itp)
 end
 
--- a/src/resizing.jl
+++ b/src/resizing.jl
@@ -227,7 +227,7 @@ function restrict_indices(r::UnitRange)
 end
 
 # imresize
-imresize(original::AbstractArray, dim1::T, dimN::T...; kwargs...) where T<:Union{Integer,AbstractUnitRange} = imresize(original, (dim1,dimN...))
+imresize(original::AbstractArray, dim1::T, dimN::T...; kwargs...) where T<:Union{Integer,AbstractUnitRange} = imresize(original, (dim1,dimN...); kwargs...)
 function imresize(original::AbstractArray; ratio, kwargs...)
     all(ratio .> 0) || throw(ArgumentError("ratio $ratio should be positive"))
     new_size = ceil.(Int, size(original) .* ratio) # use ceil to avoid 0
@@ -239,6 +239,10 @@ function imresize(original::AbstractArray, short_size::Union{Indices{M},Dims{M}}
     len_short > ndims(original) && throw(DimensionMismatch("$short_size has too many dimensions for a $(ndims(original))-dimensional array"))
     new_size = ntuple(i -> (i > len_short ? odims(original, i, short_size) : short_size[i]), ndims(original))
     imresize(original, new_size; kwargs...)
+end
+
+function imresize(original::AbstractArray, itp::Union{Interpolations.Degree,Interpolations.InterpolationType}, args...; kwargs...)
+    throw(ArgumentError("Argument support for interpolation is not supported. Are you looking for the method keyword to pass an interpolation method?"))
 end
 
 odims(original, i, short_size::Tuple{Integer,Vararg{Integer}}) = size(original, i)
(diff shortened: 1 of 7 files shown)


src/resizing.jl
```diff
@@ -262,7 +263,7 @@ julia> imresize(img, (128, 128)) # 128*128
 julia> imresize(img, (1:128, 1:128)) # 128*128
 julia> imresize(img, (1:128, )) # 128*256
 julia> imresize(img, 128) # 128*256
-julia> imresize(img, ratio = 0.5) # 
+julia> imresize(img, ratio = 0.5) #
```
johnnychen94, line 266: This seems to be a mistake in previous PRs. Could you help add the size information here?

src/resizing.jl
```diff
@@ -292,6 +293,21 @@ function imresize(original::AbstractArray{T,N}, new_size::Dims{N}) where {T,N}
...
+function imresize(original::AbstractArray{T,N}, new_size::Dims{N}, aitp::Interpolations.Lanczos4OpenCV) where {T,N}
+    Tnew = imresize_type(first(original))
+    inds = axes(original)
+    if map(length, inds) == new_size
+        dest = similar(original, Tnew, new_size)
+        if axes(dest) == inds
+            copyto!(dest, original)
+        else
+            copyto!(dest, CartesianIndices(axes(dest)), original, CartesianIndices(inds))
+        end
```
johnnychen94, line 305: I think `OffsetArrays` now preserves axes when you simply do `copy(original)`. Any specific reason for this manual written if branch?

src/resizing.jl
```diff
@@ -317,6 +333,12 @@ function imresize!(resized::AbstractArray{T,N}, original::AbstractArray{S,N}) wh
     imresize!(resized, itp)
 end
 
+function imresize!(resized::AbstractArray{T,N}, original::AbstractArray{S,N}, aitp::Interpolations.Lanczos4OpenCV) where {T,S,N}
+    # FIXME: avoid allocation for interpolation
+    itp = interpolate(original, aitp)
```
johnnychen94, line 338: @mkitti do we have a lazy version of `interpolate`? Sorry that I don't know much about Interpolations' internal.

johnnychen94 (commented):
Does this require a recent version of Interpolations? If so, then we also need to update the `[compat]` section in `Project.toml`.

> Please advise if additions to test or documentation are desired.

Yes please. Every non-trivial PR should include a test to cover the code, and every new feature should include docstrings and perhaps examples to let users know how to it.

src/resizing.jl
andrew-saydjari: Sure. Added in the second commit. Just 128*128.

src/resizing.jl
andrew-saydjari: Not sure if it is required but it mirrors the 282:295 without the Lanczos interpolation so I just copied this previous method exactly to be safe.

src/resizing.jl
mkitti: `interpolate` is pretty lazy, but the details depend on a specific interpolation kernel.

mkitti:
Why do this only for `Interpolations.Lanczos4OpenCV`? You could use `Interpolations.AbstractLanczos`.

andrew-saydjari:
@mkitti I missed that. I have updated it to AbstractLanczos. Thanks.

@johnnychen94 I added to the doc strings for both cases and think the coverage should be fixed. How do I update codecov to look at the newest commit?

I also do not know what min version of Interpolations.jl is required. Is there an easy way to check that? I suspect Lanczos has been implemented there for a while, but I will check if you tell me how.

andrew-saydjari:
Is it easy to see what test hits line 290 in the last commit? It would help me design a test to hit 305. I was hoping this might be in the detailed codecov report I cannot see.

src/resizing.jl
```diff
@@ -317,6 +334,12 @@ function imresize!(resized::AbstractArray{T,N}, original::AbstractArray{S,N}) wh
     imresize!(resized, itp)
 end
```
johnnychen94, line 335: Instead of adding more and more specifications, we could instead generalize our `imresize` API (actually you're already doing it) to pass extra args to `imresize!`, and let `imresize!` decide how to deal with it.

Unfortunately, we can't make `method` a positional argument because there's a hard-to-solve method ambiguity on `imresize(img, sizes...)`. Hence what I'm proposing is to introduce a keyword `method`:

- `imresize(img, args...; method=BSpline(Linear()))`
- `imresize!(img, args...; method=BSpline(Linear()))`.

```diff
- function imresize(original::AbstractArray{T,N}, new_size::Dims{N}) where {T,N}
+ function imresize(original::AbstractArray{T,N}, new_size::Dims{N}; kwargs...) where {T,N}
    Tnew = imresize_type(first(original))
    inds = axes(original)
    if map(length, inds) == new_size
        dest = similar(original, Tnew, new_size)
        if axes(dest) == inds
            copyto!(dest, original)
        else
            copyto!(dest, CartesianIndices(axes(dest)), original, CartesianIndices(inds))
        end
    else
-        imresize!(similar(original, Tnew, new_size), original)
+        imresize!(similar(original, Tnew, new_size), original; kwargs...)
    end
end

- function imresize!(resized::AbstractArray{T,N}, original::AbstractArray{S,N}) where {T,S,N}
+ function imresize!(resized::AbstractArray{T,N}, original::AbstractArray{S,N}; method::InterpolationType = BSpline(Linear())) where {T,S,N}
    # FIXME: avoid allocation for interpolation
-    itp = interpolate(original, BSpline(Linear()))
+   itp = interpolate(original, method)
    imresize!(resized, itp)
end
    # FIXME: avoid allocation for interpolation
    imresize!(resized, interpolate(original, method))
end
```

This way you don't need to copy-paste the existing methods. I didn't verify if this works out of the box, may I count on you to do this for other `imresize` and `imresize!` methods?

test/resizing.jl
```diff
@@ -144,5 +144,19 @@ end
             @test ndims(R) == 0
             @test !(R === A)
         end
+
+    @testset "Interpolation" begin
+        img = rand(16,16)
+        out = ImageTransformations.imresize(img,(128,128),Lanczos4OpenCV())
+        @test size(out) == (128,128)
```
johnnychen94, line 151: > Is it easy to see what test hits line 290 in the last commit? It would help me design a test to hit 305. I was hoping this might be in the detailed codecov report I cannot see.

Also need to test two more cases:

- when `img` has offset axes: `img = OffsetArray(rand(16, 16), -1, -1)` (a zero-indexed array) -- this is where the if-branch in 305 hits.
- when `sizes` are axes: `imresize(img, (0:127, 0:127), Lanczos4OpenCV())` -- this is the `Indices` method.

johnnychen94 (commented):
> I also do not know what min version of Interpolations.jl is required. Is there an easy way to check that? I suspect Lanczos has been implemented there for a while, but I will check if you tell me how.

According to https://github.com/JuliaMath/Interpolations.jl/releases (thanks to @mkitti's well-written release notes), it's v0.13.0 that first introduces Lancoz, so we need to set the compat to `0.13` only

src/resizing.jl
johnnychen94: We could also introduce `method` to `imrotate`, too, and that would just solve #111 if everything works.

andrew-saydjari:
> > I also do not know what min version of Interpolations.jl is required. Is there an easy way to check that? I suspect Lanczos has been implemented there for a while, but I will check if you tell me how.
> 
> According to https://github.com/JuliaMath/Interpolations.jl/releases (thanks to @mkitti's well-written release notes), it's v0.13.0 that first introduces Lancoz, so we need to set the compat to `0.13` only

Is that an acceptable restriction? I have implemented it, but I just want to make sure the loss of compatibility is ok.

johnnychen94:
Yes, Pkg now is quite smart to figure things out. Also, the old ImageTransformations versions are still around if user requires Interpolations < 0.13.

src/resizing.jl
andrew-saydjari: The main reason I did not do this was to maintain compatibility with the current form where for imrotate for example, you can pass Linear(), Cubic(OnGrid()), etc to degree. This becomes superfluous in the presence of method, but I suspect we should keep it for back compatibility. Thoughts on this?

johnnychen94:
> Yes, Pkg now is quite smart to figure things out. Also, the old ImageTransformations versions are still around if user requires Interpolations < 0.13.

Oh wait. If we introduce the method keyword and don't explicitly use Lancoz in this PR, then we actually don't need to restrict the compatibility for Interpolations.

src/resizing.jl
johnnychen94: The main reason we made this a positional argument is because Julia was not performant dealing with keyword arguments. Now that since keyword argument overhead is optimized away, we could switch to keyword argument.

As you may have noticed, positional arguments make method dispatching quite complicated to reason about and extend.

Whether we want to keep backward compatibility totally depends on you; we could bump a minor version if you don't want to spend that much time dealing with it. If you choose to keep backward compatibility, which is great, you then need to add some if-branches to check values, and manually throw deprecation warnings. An example of such can be found in [Documenter](https://github.com/JuliaDocs/Documenter.jl/blob/b5356729147f269d8fc8cb3412f7630b8edd3bf3/src/Documenter.jl#L358-L377)

andrew-saydjari: An intermediate option I want to propose is just to keep the ``degree`` versions of each function and add ``kwargs...`` so that the ``method`` kwarg is overridden by the existence of a degree keyword. The only problem with that for imrotate is the function on line 27 of imresizing.jl. Can you explain better why that case cannot be taken care of by just setting the default behavior for when only an array is passed to be a simple linear interpolation? Why is there so much type handling going on here?

johnnychen94: > The only problem with that for imrotate is the function on line 27 of imresizing.jl. Can you explain better why that case cannot be taken care of by just setting the default behavior for when only an array is passed to be a simple linear interpolation? Why is there so much type handling going on here?

There's no `imresizing.jl` here... Hmmm, which method are you asking?

johnnychen94: > An intermediate option I want to propose is just to keep the degree versions of each function and add kwargs... so that the method kwarg is overridden by the existence of a degree argument.

Could you give an example of your proposed API? Is it this?

```julia
imresize(img, sizes...; degree = Linear(), method = BSpline(Linear()))
```

It will be a little bit mind puzzling which keyword argument has higher priority, so what about:

```julia
imresize(img, sizes...; method = Linear())
```

with `method` accept both `InterpolationType` and `Degree`? When passing a degree, it is assumed to be BSpine.

andrew-saydjari: > > The only problem with that for imrotate is the function on line 27 of imresizing.jl. Can you explain better why that case cannot be taken care of by just setting the default behavior for when only an array is passed to be a simple linear interpolation? Why is there so much type handling going on here?
> 
> There's no `imresizing.jl` here... Hmmm, which method are you asking?

My bad, I meant interpolations.jl

andrew-saydjari: My proposed API is the following for the box_extrapolation which underlies the imrotate. Something similar for resize.

``box_extrapolation(parent::AbstractArray, args...; method::Interpolations.InterpolationType=BSpline(Linear()), kwargs...)``

but if you pass a degree argument you get the original function definition which ignores the kwarg.

``box_extrapolation(parent::AbstractArray{T,N}, degree::Interpolations.Degree, args...; kwargs...)``

The only part I am not sure this handles gracefully is the following

``box_extrapolation(parent::AbstractArray, degree::D = Linear(), args...; kwargs...) where D<:Union{Linear,Constant}``

andrew-saydjari: It seems this does a nice job of preserving ``===`` in the tests for linear and constant interpolations which are broken in my version without a replacement for that last function...

johnnychen94: I have no idea, either. My best guess is that it deals with cases such as `imresize(img, fill = 0)` where `img` is a colorant array.

This part of the code was originally written by @Evizero even before Julia 0.7 so there're a lot of decisions that get outdated and unnecessary now. I would suggest rewriting `box_extrapolation` for cleaner logic and take advantage of the keyword arguments. @Evizero has written a very verbose set of test cases so rewriting it should be pretty safe.

My previous argument on `imresize` should also be valid for `box_extrapolation`:

```julia
# method can either be `Degree` or `InterpolationType`; degree is automatically wrapped with `BSpine`.
box_extrapolation(parent::AbstractArray{T,N}; method= Linear(), fill = _default_fill(T)) where {T, N}
```

My guess is that we don't need `box_extrapolation` for either `AbstractInterpolation` and `AbstractExtrapolation`. These two methods are implementation details and can thus be removed; if you search `box_extrapolation` in other files, you'll see all calls to `box_extrapolation` are `AbstractArray`s.

andrew-saydjari:
I think I struck a good medium between our suggestions. 

For imresize, you have the usual behavior of bilinear interpolation as a default with all of the shape handling untouched because we are just passing the method as a kwarg. The API for changing the interpolation is simply

```julia
imresize!(img ;method::Interpolations.InterpolationType=BSpline(Linear()))
```

This is easier than the case of imrotate where there was previous support for ``degree`` options. For imrotate, we choose the following.

```julia
imrotate(img::AbstractArray{T}, θ::Real, args...; kwargs...)

box_extrapolation(parent::AbstractArray, args...; method::Union{Interpolations.Degree,Interpolations.InterpolationType}=Linear(), kwargs...)
```

where kwargs passes the ``method`` option through to the final box_extrapolation. The box_extrapolation has a few cases to deal with the simple ``Linear()`` and ``Constant()`` cases. There is an if statement to catch the previously supported ``degree`` argument in a way which I think is seamless. 

The only issue is that two imresize functions line 282 and 297 now throw an ambiguity error that I suppressed. Could you uncomment line 7 of runtests.jl and see if you understand why this is occurring. The only change I can see is that I added ``kwarg`` support to them. 

With that fixed, I think we will be good to go. I did use Lanczos4OpenCV() in the tests; I wanted to ensure that the newer complicated interpolations worked. I would hope that would not require a change in the compat?

src/resizing.jl
Evizero: I am hazy on the details, but I know I used to be quite fond of language-version specific workarounds to get the most out of the code. I would not be surprised if whatever motivated my designed choices has long been resolved on a language level.

So from my perspective feel free to question my code and rewrite it in a way that it makes sense today.
Think outside the `box_` :-)

andrew-saydjari:
@johnnychen94 Could I get some feedback if you have a second? I seem to be passing all the tests (see 680c6c8) except for this one ambiguity flag that I do not understand.

src/resizing.jl
```diff
@@ -227,19 +227,24 @@ function restrict_indices(r::UnitRange)
...
+function imresize(original::AbstractArray, short_size::Union{Indices{M},Dims{M}}; kwargs...) where M
     len_short = length(short_size)
     len_short > ndims(original) && throw(DimensionMismatch("$short_size has too many dimensions for a $(ndims(original))-dimensional array"))
     new_size = ntuple(i -> (i > len_short ? odims(original, i, short_size) : short_size[i]), ndims(original))
-    imresize(original, new_size)
+    imresize(original, new_size; kwargs...)
 end
+
+function imresize(original::AbstractArray, itp::Union{Interpolations.Degree,Interpolations.InterpolationType}, args...; kwargs...)
+    throw(ArgumentError("Argument support for interpolation is not supported. Are you looking for the method keyword to pass an interpolation method?"))
```
johnnychen94, line 245:
```suggestion
throw(ArgumentError("Argument support for interpolation is not supported. Are you looking for the `method` keyword to pass an interpolation method?"))
```

src/resizing.jl
```diff
@@ -277,7 +283,7 @@ function imresize(original::AbstractArray{T,0}, new_inds::Tuple{}) where T
     copyto!(similar(original, Tnew), original)
```
johnnychen94, line 283: This fixes the method ambiguity.

```diff
- function imresize(original::AbstractArray{T,0}, new_inds::Tuple{}) where T
+ function imresize(original::AbstractArray{T,0}, new_inds::Tuple{}; kwargs...) where T
```

src/resizing.jl
```diff
@@ -311,9 +317,9 @@ imresize_type(c::Gray) = Gray{imresize_type(gray(c))}
 imresize_type(c::FixedPoint) = typeof(c)
 imresize_type(c) = typeof((c*1)/1)
 
-function imresize!(resized::AbstractArray{T,N}, original::AbstractArray{S,N}) where {T,S,N}
+function imresize!(resized::AbstractArray{T,N}, original::AbstractArray{S,N}; method::Interpolations.InterpolationType=BSpline(Linear()), kwargs...) where {T,S,N}
     # FIXME: avoid allocation for interpolation
-    itp = interpolate(original, BSpline(Linear()))
+    itp = interpolate(original, method)
```
johnnychen94, line 322: So here we need a helper function:

```julia
@inline maybe_wrap_with_BSpline(itp::Interpolations.InterpolationType) = itp
@inline maybe_wrap_with_BSpline(degree::Interpolations.Degree) = BSpline(degree)
```

to make sure `imresize!` correctly understands `Linear()`

```suggestion
    itp = interpolate(original, maybe_wrap_with_BSpline(method))
```

johnnychen94 (commented):
Sorry for the delay!

src/resizing.jl
andrew-saydjari: I see, I was not considering that a case but I can certainly add that in.

andrew-saydjari: Since I have specified that the method be ``Interpoaltions.InterpolationType`` I think that fails because ``Interpolations.Linear<:Interpolations.InterpolationType`` fails since it is ``Degree``

andrew-saydjari: Nevermind, I just unioned the type.

andrew-saydjari:
Thanks for the pointers! I updated with all of the suggested changes and modified the doc strings. Let me know if there is anything else preventing a merge.

test/resizing.jl
```diff
@@ -144,5 +144,20 @@ end
...
+
+        out = ImageTransformations.imresize(img,(128,128),method=Linear())
+        @test size(out) == (128,128)
+
+        out = ImageTransformations.imresize(img,(16,16),method=Lanczos4OpenCV())
+        @test size(out) == (16,16)
+
+        @test_throws ArgumentError imresize(img, BSpline(Linear()))
+        @test_throws ArgumentError imresize(img, Linear())
+        end
```
johnnychen94, line 161: Let's also add the following test cases with different `method` choices.

- offset array and make sure the offset is correctly handled, e.g., `imresize(img, (128, 128)) == imresize(OffsetArray(img, -1, -1), (128, 128))`
- axes: `imresize(img, (0:127, 0:127))` and make sure the `axes(out) == (0:127, 0:127)` and `OffsetArrays.no_offset_view(out) == imresize(img, (128, 128))`

test/resizing.jl
```diff
@@ -144,5 +144,20 @@ end
             @test ndims(R) == 0
             @test !(R === A)
         end
+
+    @testset "Interpolation" begin
+        img = rand(16,16)
+        out = ImageTransformations.imresize(img,(128,128),method=Lanczos4OpenCV())
+        @test size(out) == (128,128)
+
+        out = ImageTransformations.imresize(img,(128,128),method=Linear())
```
johnnychen94, line 153: We need to test the default behavior:

```julia
imresize(img, (128,128), method=Linear()) == imresize(img, (128,128))
```

`imresize` is exported so no need to write `ImageTransformations.imresize`.

johnnychen94 (approved):
There're perhaps some more details that could get improved but this PR already did a lot of things! :tada:

The last change request is about the tests; I've left a comment in `imresize`, we should also do it for `imrotate` and `warp`. That is, we also need to add tests for `imrotate` and `warp` to make sure `Linear()`, `BSpline(Linear())` and `Lancoz4OpenCV()` are correctly handled.

As you may have noticed, `box_extrapolation` is a little bit hard to intuitive about, so after this PR gets merged, I'll submit a refactor PR to simplify it, and then make a release.

test/resizing.jl
johnnychen94: Format isn't a critical issue but also worths noting:

```diff
- imresize(img, (128,128), method=Linear())
+ imresize(img, (128,128), method=Linear())
```

See also the blue style: https://github.com/invenia/BlueStyle#whitespace

test/resizing.jl
andrew-saydjari: Are you suggesting these for all 3 interpolation methods `Linear`, `BSpline(Linear())`, and `Lanczos4OpenCV()`?

andrew-saydjari: ok, done for imresize. I tried to copy the formats in the imrotate and warp sections for the tests. Let me know if there are more that should be done.

johnnychen94 (approved):

johnnychen94:
@andrew-saydjari thanks for the great PR! :tada: It has solved so many legacy issues.

johnnychen94:
@andrew-saydjari It's not easy work to do #116 so I just tagged v0.8.11 for this PR.

Issue #64, opened 2019-01-11
Can warp also be used as a remap like in opencv?


jc211:
Is it possible/recommended to add an overload to warp that takes a lookup table instead of a transform. The functionality would be similar to OpenCVs remap found [here](https://docs.opencv.org/2.4/modules/imgproc/doc/geometric_transformations.html?highlight=remap#void%20remap(InputArray%20src,%20OutputArray%20dst,%20InputArray%20map1,%20InputArray%20map2,%20int%20interpolation,%20int%20borderMode,%20const%20Scalar&%20borderValue)).

PR #131, opened 2021-07-12
add documentation page

johnnychen94:
Three pages are added:

- index page: package overview and a simple explanation on the warp internal
- function reference page: organized in three parts: high-level, low-level API, and utilities.
- demo page: currently only add swirl example to illustrate how to use custom warp map (address #64)

Irrelevant change to unit test CI:

- only test Julia:1 on windows and macos platform because of the limited CI resource
- remove cron job because GHA would otherwise disable the CI after 30(or 60) days inactivity.


johnnychen94:
There seems a breaking Interpolations version introduced; I'll try to fix it in a new PR.

PR #136, opened 2021-07-13
remove all summary test

johnnychen94:
`summary` results are not very stable; it differs from version to version. Hence I replace all `summary` tests with `axes`, `eltype` and `size` tests.


Evizero:
I agree wit your point. the one thing I would argue is that it would be nice to at least test that the summary code can be executed without an error

johnnychen94:
I've added the `summary` back with `@test_nowarn summary(...)`. But actually, since ImageTransformations doesn't override `summary` method for any of the view types, the `summary(...)` should just work.

PR #138, opened 2021-07-18
deprecate `invwarpedview` in favor of `InvWarpedView`

johnnychen94:
Should be the last PR before v0.9 release.

Now that `InvWarpedView` accepts `method` and `fillvalue` keywords, and also commits to the `parent(wv) === img` assumption. (This is also an improvement to https://github.com/JuliaImages/ImageTransformations.jl/pull/24 and thus makes `invwarpedview` not really necessary.)

This also introduces the same set of already deprecated API to keep consistent with `warpedview`, e.g., `InvWarpedView(img, tinv, inds, method, fillvalue)` is newly added as a deprecated method.

Other unrelated changes:
  - propagate `@inbounds` for `getindex` method for both `WarpedView` and `InvWarpedView`
  - simplify `Base.size` methods and remove `IndexStyle`; it's sufficient to use the fallback implementation in Base.

TODO:

- [x] bring the `SubArray` trick back appropriately (This trick is directly removed)


timholy:
This seems reasonable. One distinction between `InvWarpedView` and `invwarpedview` is that the first *should* be a constructor and always return an `InvWarpedView`. The second would have license to, e.g., strip a `WarpedView` wrapper that happens to be inverted from the one being applied. That's probably a corner case that would only arise by design rather than in "real" usage.

johnnychen94:
The more concerning issue I find with this trick with this "license" to merge two or more lazy warp views through `SubArray` is that out-of-domain values are not guaranteed to be valid. This can be a deeply buried root of potential bugs so I choose to remove this to make things easier to intuit.

```julia
img_camera = testimage("cameraman")
img = similar(img_camera) # uninitialized array
v = view(img, 75:195, 245:370)
v .= img_camera[v.indices...]

tfm = recenter(RotMatrix(-pi/8), center(img_camera))
wrong_result = collect(invwarpedview(v, tfm)) # there's no way to check if `parent(v)` are filled with real useful data.
correct_result = collect(InvWarpedView(v, tfm))
mosaic(wrong_result, correct_result; nrow=1)
```

image

It should be the caller's responsibility to check whether this optimization is valid and not the callee `invwarpedview`/`warpedview`.

timholy:
I like this way of thinking! 100% agreed.

PR #142, opened 2021-07-27
[WIP] fixed point perspective for `imresize`, `zoom` and `imrotate`

johnnychen94:
Let me recite [our newly-added documentation](https://juliaimages.org/ImageTransformations.jl/dev/#index_image_warping) for the symbols and terminology:

A (backward-mode) warp operation consists of two parts: the value estimator τ (which we always use Interpolations.jl in this package), and the backward coordinate map ϕ. For some special warping operations such as scale(`imresize`), zoom, and rotation: unless it's an identity map, there exists one and only one point that satisfies `ϕ(p) == p`, and this `p` is defined as the fixed point.

This PR aims to provide consistent reasoning on this "fixed point" concept for these three operations.

### `imresize`

Because `imresize` always spans the entire source domain, i.e., `CartesianIndices(src_img)`, exposing a control keyword for fixed point only changes the offset of the output. For example: `imresize(rand(10, 10), (5, 5); fixed_point = (5, 5))` requires the output array (an offset array) axes be `(3:7, 3:7)` so that `ϕ((5, 5)) == (5, 5)`.

Also note that `imresize(rand(10, 10), (5, 5))` returns an `Array` in previous ImageTransformations versions, and I don't think we should change this because 1) it exists for so long and `Array` as the base julia type has the best support among the entire ecosystem, and 2) `OffsetArray` introduces an extra step when computing the indices, so it might hurt the performance (although in many cases that I've observed, Julia compiler just optimizes this overhead away).

Thus for type stability, we can't make `fixed_point` a keyword, and this is why we did `CenterPoint` thing in #129 and it works pretty well.

However, one key question that #129 doesn't answer is: what is the default fixed point? And trying to answer this leads me to open a new PR here.

The current master version of ImageTransformation doesn't have it well-defined: for `Array` it is `(1, 1)`. But for `OffsetArray` it's quite inconsistant because it maps `p0 = map(first, axes(offset_img))` to `q0 = (1, 1)`, which says the fixed point is neither `(1, 1)` or `p0`.

The first commit of this PR 605be80 does two things:

- it fixes the type instability by unconditionally make the `imresize(img, inds)` method output `OffsetArray`.
- (__breaking change__) the fixed point defaults to `map(first, axes(img))`: `imresize(offset_img, ...)` now outputs an `OffsetArray` instead of an `Array`. This is no doubt a breaking change but it's for a more consistent result when speaking of the fixed point things.

To support #129, we then just need some manual axes shifts on the output:

```julia
img = testimage("cameraman")
fixed_point = OffsetArrays.center(img)

o = @. fixed_point ÷ 2 + 1
out = imresize(img; ratio=0.5, method=Constant()) # we use nearest interpolation to avoid interpolation on grid point
outo = OffsetArray(out, OffsetArrays.Origin(o))
outo[fixed_point...] == img[fixed_point...] # true
```

I'm now unsure if we should support fixed point concept for `imresize` because it is really just a matter of offset shift.

### [WIP] `zoom`

`zoom` works quite similar to `imresize` except that the canvas size isn't changed, and the fixed point concept plays a key role in zoom operation.

The second commit 33b5559 is a WIP for this new high-level API.

The currently implemented API is `zoom(img; ratio, [fixed_point], [method], [fillvalue])`, I might also want to introduce `zoom(img, size_or_indices; [fixed_point], kwargs...)` but I still need to think about if it's possible to interpret the `indices` input.

Top-left point as fixed point : `map(first, axes(img))`

```julia
img = imresize(testimage("camera"), 128, 128)
ImageShow.gif([zoom(img; ratio=r, fixed_point=(1, 1)) for r in 0.5:0.1:2])
```

topleft

Center point as fixed point: `OffsetArrays.center(img)`

```julia
img = imresize(testimage("camera"), 128, 128)
ImageShow.gif([zoom(img; ratio=r, fixed_point=(64, 64)) for r in 0.5:0.1:2])
```

center

### [TODO] `imrotate`

The proposed API is to add `fixed_point` keyword.

---

Todo:

- [ ] fix method ambiguities for `imresize`
- [ ] add `zoom`
- [ ] add keyword `fixed_point` for `imrotate`

---

cc: @ginkulv

### DIFF
--- a/src/resizing.jl
+++ b/src/resizing.jl
@@ -6,7 +6,23 @@ function restrict(img::Union{WarpedView, InvWarpedView}, dims::Integer)
     restrict(OffsetArray(collect(img), axes(img)), dims)
 end
 
-# imresize
+###########
+# imresize/imresize!
+#
+# `imresize` API:
+#   - `imresize(original; ratio, [method])`
+#   - `imresize(original, inds; [method])`
+#   - `imresize(original, sz; [method])`
+# The output type:
+#   - If `inds` method get triggered, the output will unconditionally be `OffsetArray` type.
+#   - Otherwise, it should preserve the original input array type.
+#
+# `imresize!` API:
+#   - `imresize!(resized, original::AbstractArray; [method])`
+#   - `imresize!(resized, original::AbstractInterpolation)`
+###########
+
+
 imresize(original::AbstractArray, dim1::T, dimN::T...; kwargs...) where T<:Union{Integer,AbstractUnitRange} = imresize(original, (dim1,dimN...); kwargs...)
 function imresize(original::AbstractArray; ratio, kwargs...)
     all(ratio .> 0) || throw(ArgumentError("ratio $ratio should be positive"))
@@ -26,16 +42,18 @@ function imresize(original::AbstractArray, itp::Union{Interpolations.Degree,Inte
 end
 
 odims(original, i, short_size::Tuple{Integer,Vararg{Integer}}) = size(original, i)
-odims(original, i, short_size::Tuple{}) = axes(original, i)
+odims(original, i, short_size::Tuple{}) = size(original, i)
 odims(original, i, short_size) = oftype(first(short_size), axes(original, i))
 
 """
     imresize(img, sz; [method]) -> imgr
-    imresize(img, inds; [method]) -> imgr
+    imresize(img, inds; [method]) -> imgr::OffsetArray
     imresize(img; ratio, [method]) -> imgr
 
-upsample/downsample the image `img` to a given size `sz` or axes `inds` using interpolations. If
-`ratio` is provided, the output size is then `ceil(Int, size(img).*ratio)`.
+upsample/downsample the image `img` to a given size `sz` or axes `inds` using interpolations.
+
+If `ratio` is provided, the output size is then `ceil(Int, size(img).*ratio)`. If axes information is
+provided by passing `inds`, then the output array is `OffsetArray`.
 
 !!! tip
     This interpolates the values at sub-pixel locations. If you are shrinking the image, you risk
@@ -104,6 +122,8 @@ function imresize(original::AbstractArray{T,N}, new_size::Dims{N}; kwargs...) wh
         if axes(dest) == inds
             copyto!(dest, original)
         else
+            # Non 1-based case as a fallback solution
+            # The OffsetArray case is specially handled to also output OffsetArray
             copyto!(dest, CartesianIndices(axes(dest)), original, CartesianIndices(inds))
         end
     else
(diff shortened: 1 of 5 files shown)


src/resizing.jl
```diff
@@ -104,19 +122,42 @@ function imresize(original::AbstractArray{T,N}, new_size::Dims{N}; kwargs...) wh
...
+            # Non 1-based case as a fallback solution
+            # The OffsetArray case is specially handled to also output OffsetArray
             copyto!(dest, CartesianIndices(axes(dest)), original, CartesianIndices(inds))
         end
     else
         imresize!(similar(original, Tnew, new_size), original; kwargs...)
     end
 end
 
+function imresize(original::OffsetArray{T,N}, new_size::Dims{N}; kwargs...) where {T,N}
```
timholy, line 134: OffsetArray is not necessarily the only array type that doesn't have indexing starting with 1; there are several other examples. What's the reason for the special method? It looks almost identical to the above.

src/resizing.jl
```diff
@@ -104,19 +122,42 @@ function imresize(original::AbstractArray{T,N}, new_size::Dims{N}; kwargs...) wh
...
+end
+
 function imresize(original::AbstractArray{T,N}, new_inds::Indices{N}; kwargs...) where {T,N}
     Tnew = imresize_type(first(original))
+    # The pirated `similar` method from OffsetArrays will be triggered
+    # thus for type stability, we unconditionally output an `OffsetArray`.
+    origin = OffsetArrays.Origin(map(first, new_inds))
     if axes(original) == new_inds
-        copyto!(similar(original, Tnew), original)
+        OffsetArray(copyto!(similar(original, Tnew), original), origin)
```
timholy, line 158: Rather than manually create an `OffsetArray`, how about use `similar(original, Tnew, new_inds)`? That will create an OffsetArray if `new_inds[i] isa UnitRange` rather than a `Base.OneTo`.

test/resizing.jl
```diff
@@ -25,28 +25,62 @@ end
...
+            @test test_imresize_interface(img, (5,10), (1:5, 1:10), 5) isa Array
+            @test test_imresize_interface(img, (5,10), (1:5, 1:10), (5,)) isa Array
+
+            @test test_imresize_interface(img, (5,5), (1:5, 1:5), ratio = 0.5) isa Array
+            @test test_imresize_interface(img, (20,20), (1:20, 1:20), ratio = 2) isa Array
+            @test test_imresize_interface(img, (20,20), (1:20, 1:20), ratio = (2, 2)) isa Array
+            @test test_imresize_interface(img, (20,10), (1:20, 1:10), ratio = (2, 1)) isa Array
+            
+            # indices method always return OffsetArray
+            @test test_imresize_interface(img, (5,5), (1:5, 1:5), (1:5,1:5)) isa OffsetArray
```
timholy, line 51: Also add
```julia
@test test_imresize_interface(img, (5,5), (1:5, 1:5), (Base.OneTo(5),Base.OneTo(5))) isa Array
```

timholy (commented):
LGTM. I didn't check the zoom calculations, but the idea seems nice.

timholy:
How about this as a proposal:
- `imresize(img, sz::Dims)` is for the "I live in a 1s-based world, don't worry about the indices". This always returns an `Array` and throws an error if `img` has axes that don't start at 1.
- `imresize(img, sz::Dims, fp::Dims)` is for a fixed point and always returns an array whose type is dictated by appropriate computations on the axes of `img`. If they are `Base.OneTo` then we switch to `UnitRange` and get an OffsetArray, but if it's, e.g., a [`CatIndices.URange`](https://github.com/JuliaArrays/CatIndices.jl/blob/master/src/bidirectional.jl) then it creates that output type.

johnnychen94:
This PR becomes more delicate than I expected, and it's quite hard to find a way out for `imresize`. Thus I've converted this into a draft, and plan to split this PR into smaller parts so that it's easier to review and discuss. For instance, 1) adding `zoom` and 2) add `fixed_point` keyword for `imrotate` should be non-controversial.

PR #143, opened 2021-07-27
Restrict rounding to the computed corners

timholy:
Rounding the transformation makes assumptions about the nature of
these transformations. However, in principle any kind of function
mapping points to new points should be supported.  It seems safer to
perform rounding on the *result* of the transformation.

This will require https://github.com/JuliaMath/Interpolations.jl/pull/451 to pass the Lanczos tests.

The other test failure is because we're testing "fill" boundary conditions and without rounding the result is just barely beyond-the-edge. Any thoughts on how you want to approach that? One option would be to slightly round the coordinates before interpolation; a reasonably fast way to do that might be

```julia
__round(x::Float64) = Float64(Float32(x))
__round(x::Float32) = Float32(Float16(x))    # surprisingly, this is ~1ns also
```

but that will trigger a cascade of other small precision errors.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -19,13 +19,14 @@ ColorVectorSpace = "0.6, 0.7, 0.8, 0.9"
 CoordinateTransformations = "0.5, 0.6"
 ImageBase = "0.1.1"
 ImageCore = "0.8.1, 0.9"
-Interpolations = "0.9, 0.10, 0.11, 0.12, 0.13"
+Interpolations = "0.13.4"
 OffsetArrays = "0.10, 0.11, 1.0.1"
 Rotations = "0.12, 0.13, 1.0"
 StaticArrays = "0.10, 0.11, 0.12, 1.0"
 julia = "1"
 
 [extras]
+EndpointRanges = "340492b5-2a47-5f55-813d-aca7ddf97656"
 ImageMagick = "6218d12a-5da1-5696-b52f-db25d2ecc6d1"
 LinearAlgebra = "37e2e46d-f89d-539d-b4ee-838fcccc9c8e"
 ReferenceTests = "324d217c-45ce-50fc-942e-d289b448e8cf"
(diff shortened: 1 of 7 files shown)


johnnychen94 (commented):
This in general is a better patch than my previous https://github.com/JuliaImages/ImageTransformations.jl/pull/107

> One option would be to slightly round the coordinates before interpolation ... but that will trigger a cascade of other small precision errors.

I'm supportive of this. Since we're planning a big and breaking ImageTransformation v0.9 release, the small precision errors can be expected :) That said, fixing the ImageTransformation tests seems sufficient for this PR IMO.

> `# surprisingly, this is ~1ns also

Is this because `Float16` in Julia is implemented in software?

timholy:
> >   One option would be to slightly round the coordinates before interpolation
> I'm supportive of this...That said, fixing the ImageTransformation tests seems sufficient for this PR IMO.

I like that idea---I might mark it as `@test_broken` or `@test_skip` for the meantime. It occurs to me that there's a simpler approach: define a `FilledExtrapolation` type that allows a small bit of "buffering" at the edges. The only problem basically gets caused when the evaluation point is 0.99999999... rather than 1.0 (where 1.0 is the edge of the image), and so a tiny bit of grace specifically at the edges seems like it would solve the problem while not reducing precision elsewhere. (The more I think about it, the more generic rounding would be a really bad idea for things like autodiff.)

> >    `# surprisingly, this is ~1ns also

> Is this because Float16 in Julia is implemented in software?

That's why I'm surprised, but it turns out this specific transformation comes down to CPU instructions:

```julia
julia> @code_typed __round(0.8f0)
CodeInfo(
1 ─ %1 = Base.fptrunc(Base.Float16, x)::Float16
│   %2 = Base.fpext(Base.Float32, %1)::Float32
└──      return %2
) => Float32
```

and these boil down to CPU instructions:
- https://releases.llvm.org/9.0.0/docs/LangRef.html#fptrunc-to-instruction
- https://releases.llvm.org/9.0.0/docs/LangRef.html#fpext-to-instruction

johnnychen94:
I'll finish my other two PRs in the next few days so we don't block the v0.9 release permanently.

PR #148, opened 2021-10-09
Use higher precision in imrotate

### LINKED ISSUE #145: `imrotate(img, pi/2)` changes size of image
```julia 
img = rand(RGB{N0f8}, 1000, 1000)
img2 = imrotate(img, π/2)
size(img) # (1000, 1000)
size(img2) # (1002, 1002)
```

I realize it's pretty naive to just ask "why doesn't this work as expected?". I get the sense that "simply" rotating an image is not so simple. (I know very little about image processing)

I'm wondering how I should get the desired result for rotations where the degree is multiples of `pi/2`. This might be related to #119 although I'm not sure. 

Would it make sense to just define a new imrotate specifically for these factors of `pi/2`?

### LINKED ISSUE #147: Clarification of rotations
Hi,

I have a Python script in which there is a call to `scipy.ndimage.rotate` for rotating a 2D array, similar to the following
```julia
using PyCall
ndimage = pyimport("scipy.ndimage")
# test array
img = Float64[1 2 3 4; 5 6 7 8; 9 10 11 12; 13 14 15 16]
img_90 = ndimage.rotate(img, 90, reshape=false)
```

I came across this package when searching for the same kind of operation on a plain 2D array in Julia. However, soon I got confused by the rotation methods provided here, especially `imrotate`.
```julia
using ImageTransformations
using Rotations: RotMatrix
using CoordinateTransformations: recenter

# test array
img = Float64[1 2 3 4; 5 6 7 8; 9 10 11 12; 13 14 15 16]

# define transformation
trfm = recenter(RotMatrix(pi/2), center(img))
imgw = warp(img, trfm) # this is what I expected, same as in scipy

imgr = imrotate(img, pi/2) # this gives a 6x6 OffsetArray with values that I don't quite understand
```
where `imrotate` returned
```julia
6×6 OffsetArray(::Matrix{Float64}, 0:5, 0:5) with eltype Float64 with indices 0:5×0:5:
 NaN  NaN    NaN        NaN        NaN        NaN
 NaN  NaN    NaN          4.99995  NaN        NaN
 NaN   14.0    9.99998    5.99999  NaN        NaN
 NaN  NaN     11.0        7.00002    3.00003  NaN
 NaN  NaN     12.0      NaN        NaN        NaN
 NaN  NaN    NaN        NaN        NaN        NaN
```

I tried [the example given in the doc](https://juliaimages.org/ImageTransformations.jl/dev/reference/#ImageTransformations.imrotate) for `imrotate`, and it looked just fine.

Can you kindly point out what I missed here? What is the closest possible translation from `scipy.ndimage.rotate`?

timholy:
Fixes #147
Fixes #145

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -27,6 +27,7 @@ julia = "1"
 
 [extras]
 EndpointRanges = "340492b5-2a47-5f55-813d-aca7ddf97656"
+ImageIO = "82e4d734-157c-48bb-816b-45c225c6df19"
 ImageMagick = "6218d12a-5da1-5696-b52f-db25d2ecc6d1"
 LinearAlgebra = "37e2e46d-f89d-539d-b4ee-838fcccc9c8e"
 ReferenceTests = "324d217c-45ce-50fc-942e-d289b448e8cf"
@@ -34,4 +35,4 @@ Test = "8dfed614-e22c-5e08-85e1-65c5234f0b40"
 TestImages = "5e47fb64-e119-507b-a336-dd2b206d9990"
 
 [targets]
-test = ["EndpointRanges", "ImageMagick", "LinearAlgebra", "ReferenceTests", "Test", "TestImages"]
+test = ["EndpointRanges", "ImageIO", "ImageMagick", "LinearAlgebra", "ReferenceTests", "Test", "TestImages"]


timholy:
Let me acknowledge that the broken test isn't ideal. However:
```julia
julia> mod2pi(2π)
6.283185307179586

julia> mod2pi(nextfloat(2π))
6.432490598706546e-16
```
so it cannot exactly hit 0, and hence the `NaN` pattern at the edge is different. But it's the only case from among all the angles where there is a difference.

timholy:
```diff
diff --git a/Project.toml b/Project.toml
index 7439b2a..964d5f6 100644
--- a/Project.toml
+++ b/Project.toml
@@ -28,10 +28,11 @@ julia = "1"
 [extras]
 EndpointRanges = "340492b5-2a47-5f55-813d-aca7ddf97656"
 ImageIO = "82e4d734-157c-48bb-816b-45c225c6df19"
+ImageMagick = "6218d12a-5da1-5696-b52f-db25d2ecc6d1"
 LinearAlgebra = "37e2e46d-f89d-539d-b4ee-838fcccc9c8e"
 ReferenceTests = "324d217c-45ce-50fc-942e-d289b448e8cf"
 Test = "8dfed614-e22c-5e08-85e1-65c5234f0b40"
 TestImages = "5e47fb64-e119-507b-a336-dd2b206d9990"
 
 [targets]
-test = ["EndpointRanges", "ImageIO", "LinearAlgebra", "ReferenceTests", "Test", "TestImages"]
+test = ["EndpointRanges", "ImageIO", "ImageMagick", "LinearAlgebra", "ReferenceTests", "Test", "TestImages"]

```

timholy:
Just trying to catch up on rather a long time of letting you carry the load!

PR #152, opened 2021-11-02
Add an overall module docstring

zelunw:
?ImageTransformations

Created using ModuleDocstrings.jl
with introduction and docstring for `restrict` copied from https://juliaimages.org/ImageTransformations.jl/dev/
and examples from
https://juliaimages.org/latest/pkgs/transformations/

### DIFF
--- a/src/ImageTransformations.jl
+++ b/src/ImageTransformations.jl
@@ -1,3 +1,40 @@
+"""
+
+This package provides support for image resizing, image rotation, and other spatial transformations of arrays.
+
+- `InvWarpedView`: Create a view of `img` that lazily transforms any given index `I` passed to `wv[I]` so that `wv[I] == img[inv(tinv)(I)]`.
+- `WarpedView`: Create a view of `img` that lazily transforms any given index `I` passed to `wv[I]` so that `wv[I] == img[tform(I)]`.
+- `imresize`: Upsample/downsample the image `img` to a given size `sz` or axes `inds` using interpolations.
+- `restrict`: A much more efficient version of `imresize` that two-folds/down-samples image to approximate 1/2 size. (This is now provided by ImageBase.)
+- `imrotate`: Rotate image `img` by `θ`∈[0,2π) in a clockwise direction around its center point.
+- `warp`: Transform the coordinates of `img`, returning a new `imgw` satisfying `imgw[I] = img[tform(I)]`.
+- `autorange`: For given transformation `tform`, return the "smallest" range indices that preserves all information from `A` after applying `tform`.
+
+There are in-place version of many of the functions, e.g., imresize! etc.
+
+Resize example:
+
+```julia
+using ImageTransformations, TestImages
+
+img = testimage("mandrill")
+
+img_small = imresize(img, ratio=1/8)
+img_medium = imresize(img_small, size(img_small).*2)
+```
+
+Warping example:
+
+```julia
+using ImageTransformations, TestImages, CoordinateTransformations, Rotations
+
+img = testimage("camera");
+
+# define transformation
+trfm = recenter(RotMatrix(pi/8), center(img));
+imgw = warp(img, trfm);
+```
+"""
 module ImageTransformations
 
 using ImageCore
--- a/src/ImageTransformations.jl
+++ b/src/ImageTransformations.jl
@@ -2,12 +2,12 @@
 
 This package provides support for image resizing, image rotation, and other spatial transformations of arrays.
 
-- `InvWarpedView`: Create a view of `img` that lazily transforms any given index `I` passed to `wv[I]` so that `wv[I] == img[inv(tinv)(I)]`.
 - `WarpedView`: Create a view of `img` that lazily transforms any given index `I` passed to `wv[I]` so that `wv[I] == img[tform(I)]`.
+- `warp`: An eager evaluation variant of `WarpedView`. Transform the coordinates of `img`, returning a new `imgw` satisfying `imgw[I] = img[tform(I)]`. 
+- `InvWarpedView`: Create a view of `img` that lazily transforms any given index `I` passed to `wv[I]` so that `wv[I] == img[inv(tinv)(I)]`.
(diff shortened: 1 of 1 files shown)


src/ImageTransformations.jl
```diff
@@ -1,3 +1,40 @@
+"""
+
+This package provides support for image resizing, image rotation, and other spatial transformations of arrays.
+
+- `InvWarpedView`: Create a view of `img` that lazily transforms any given index `I` passed to `wv[I]` so that `wv[I] == img[inv(tinv)(I)]`.
+- `WarpedView`: Create a view of `img` that lazily transforms any given index `I` passed to `wv[I]` so that `wv[I] == img[tform(I)]`.
```
timholy, line 6: Could you perhaps list `WarpedView` first since it's conceptually the simpler of the two?

src/ImageTransformations.jl
```diff
@@ -1,3 +1,40 @@
...
+- `imresize`: Upsample/downsample the image `img` to a given size `sz` or axes `inds` using interpolations.
+- `restrict`: A much more efficient version of `imresize` that two-folds/down-samples image to approximate 1/2 size. (This is now provided by ImageBase.)
```
timholy, line 8: Maybe we should take this one out now, since it has recently migrated to ImageBase.

src/ImageTransformations.jl
```diff
@@ -1,3 +1,40 @@
...
+- `imrotate`: Rotate image `img` by `θ`∈[0,2π) in a clockwise direction around its center point.
+- `warp`: Transform the coordinates of `img`, returning a new `imgw` satisfying `imgw[I] = img[tform(I)]`.
```
timholy, line 10: Maybe this should also appear next to `WarpedView` and explain that it's an eager variant. https://en.wikipedia.org/wiki/Lazy_evaluation

timholy (commented):
Thanks for this! With a couple of tweaks I think this is mergeable.

src/ImageTransformations.jl
johnnychen94: I think we still need this one since it's highly related to this package. So that people can know there's a better version of `imresize` without hanging around in tens of JuliaImages packages.

timholy: Good point.

zelunw: Keeping it. Edited end of docstring to 'ImageBase.restrict' for ease of search. Thanks for the input!

src/ImageTransformations.jl
```diff
@@ -1,3 +1,40 @@
...
+
+- `WarpedView`: Create a view of `img` that lazily transforms any given index `I` passed to `wv[I]` so that `wv[I] == img[tform(I)]`.
+- `warp`: An eager evaluation variant of `WarpedView`. Transform the coordinates of `img`, returning a new `imgw` satisfying `imgw[I] = img[tform(I)]`. 
+- `InvWarpedView`: Create a view of `img` that lazily transforms any given index `I` passed to `wv[I]` so that `wv[I] == img[inv(tinv)(I)]`.
+- `imresize`: Upsample/downsample the image `img` to a given size `sz` or axes `inds` using interpolations.
+- `restrict`: A much more efficient version of `imresize` that two-folds/down-samples image to approximate 1/2 size. (This is now provided by `ImageBase.restrict`)
+- `imrotate`: Rotate image `img` by `θ`∈[0,2π) in a clockwise direction around its center point.
+- `autorange`: For given transformation `tform`, return the "smallest" range indices that preserves all information from `A` after applying `tform`.
+
+There are in-place version of many of the functions, e.g., imresize! etc.
```
johnnychen94, line 13:
```suggestion
There are in-place version of many of the functions, e.g., `imresize!` etc.
```

src/ImageTransformations.jl
```diff
@@ -1,3 +1,40 @@
...
+
+```julia
+using ImageTransformations, TestImages, CoordinateTransformations, Rotations
+
+img = testimage("camera");
+
+# define transformation
+trfm = recenter(RotMatrix(pi/8), center(img));
+imgw = warp(img, trfm);
+```
```
johnnychen94, line 36: Quite curiously, can we make these examples `jldoctest` and trigger Documenter's doctest here? I'm a little bit worried that the contents get outdated in, say, 2 years. And then we get new https://github.com/JuliaImages/Images.jl/issues/982 issue to fix 😆

johnnychen94 (approved):
A typo and a question. Others look good to me. Thanks for the proposal.

src/ImageTransformations.jl
timholy: The problem is that these little hints don't generate any output. I guess we'd at least find out if they run? It would be as simple as changing ` ```julia -> ```jldoctest`.

src/ImageTransformations.jl
```diff
@@ -1,3 +1,40 @@
...
+- `imresize`: Upsample/downsample the image `img` to a given size `sz` or axes `inds` using interpolations.
+- `restrict`: A much more efficient version of `imresize` that two-folds/down-samples image to approximate 1/2 size. (This is now provided by `ImageBase.restrict`)
+- `imrotate`: Rotate image `img` by `θ`∈[0,2π) in a clockwise direction around its center point.
+- `autorange`: For given transformation `tform`, return the "smallest" range indices that preserves all information from `A` after applying `tform`.
+
+There are in-place version of many of the functions, e.g., `imresize!` etc.
+
+Resize example:
+
+```julia
```
timholy, line 17:
```suggestion
```jldoctest
```

src/ImageTransformations.jl
```diff
@@ -1,3 +1,40 @@
...
+
+img = testimage("mandrill")
+
+img_small = imresize(img, ratio=1/8)
+img_medium = imresize(img_small, size(img_small).*2)
+```
+
+Warping example:
+
+```julia
```
timholy, line 28:
```suggestion
```jldoctest
```

timholy (approved):

timholy:
Thanks, @zelunw!

PR #158, opened 2022-07-24
WIP: use StaticArraysCore

johnnychen94:
People are blaming ImageTransformations too heavy to load. One of the main latency comes from StaticArrays by checking `@time_imports` (in Julia 1.8)


johnnychen94:
Hmmm, it turns out that `Rotations` can hardly replace StaticArrays with StaticArraysCore, @hyrodium since you know more Rotations internal than me, can you help check if this is doable?

hyrodium:
`StaticArraysCore.jl` defines types but not methods for the types. So it seems difficult to replace `StaticArrays.jl` with `StaticArraysCore.jl` for the packages that `Rotations.jl` depends on...

johnnychen94:
Yeah I figured so. Let's just close this PR since it won't change anything.

Issue #165, opened 2023-11-30
Support v0.15 of Interpolations.jl


cgarling:
Interpolations.jl v0.15.0 is now out and it would be nice to update this package's compat to support it. I tried to do this on my own fork so I could submit a PR but the new version of Interpolations.jl breaks the test on line 65 of `test/interpolations.jl` and I don't have time to look into why at the moment.

roflmaostc:
Can be closed because of #168

PR #168, opened 2024-01-03
Include Interpolations.jl 0.15 support

roflmaostc:
Hi,

as raised in #165 I added support.

Not sure why this failing test was there but it makes sense to remove it?

This update is important for me since it breaks some Zygote with Julia 1.10 ultimately in the dependency chain.

Best,

Felix


roflmaostc:
Seems like this tests fails with Interpolations 0.14 but works with 0.15? Not sure what to do here

timholy:
Thanks! New release coming momentarily.

PR #173, opened 2025-05-16
Support Interpolations v0.16

cgarling:
Since your CompatHelper runs are disabled, here's a manual PR. 

I've also added dependabot.yml and given explicit permissions to the compathelper action.


cgarling:
Actions appear to be broken; I will take a shot at fixing

cgarling:
Ok I think this looks good, requesting review

johnnychen94:
@cgarling Thanks!

PR #176, opened 2026-04-22
Update to Documenter.jl v1

icweaver:
Howdy, would something like this be alright?

I found myself wanting to link to some functions here with DocumenterInterLinks.jl, but I think the docs need to be re-built with a more recent version of Documenter.jl first so that the `objects.inv` file can be created.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,3 +2,5 @@
 *.jl.*.cov
 *.jl.mem
 Manifest.toml
+docs/build
+docs/examples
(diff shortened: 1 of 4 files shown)


icweaver:
Howdy, just sending a small ping here

icweaver:
Hi, just another small ping

icweaver:
Hi all, just sending another ping

timholy:
Thanks for your patience and contributions!

PR #183, opened 2026-07-16
Add FastInterpolations.jl as an `imresize` backend

mgyoo86:
Hi! I'm the author of [FastInterpolations.jl](https://github.com/ProjectTorreyPines/FastInterpolations.jl), a new interpolation package focused on high performance.
I'm looking for more places to use it, and `imresize` seems like a good fit.

I recently added new features to FastInterpolations.jl to make it even faster for `imresize`, and I'd like to contribute it here as a new & faster backend.

It is a small wrapper around the existing `method` keyword, and all the existing methods still work as before.
I set `FI.LinearInterp()` as the new default. 
(other FastInterpolations methods are also available, i.e. FI.CubicInterp())
If you prefer to keep Interpolations as the default and make this optional, that is a one-line change.

The results match the current linear output, up to floating-point rounding error.
For fixed-point types (`N0f8`, `N0f16`), the difference is at most one quantization level.

## Benchmarks
For linear interpolation, the new FastInterpolations backend is up to ~2× faster for downsampling and ~4× for upsampling, with the same result as before.

Interpolations.jl `Linear()` vs the new default FastInterpolations.jl `LinearInterp()`, bilinear, measured in place (`imresize!` into a pre-allocated buffer). 
Speedup = Interpolations time / FastInterpolations time. 
Single-threaded, on mac M1 machine.

| Image | Operation | Output | Interpolations | FastInterpolations | Speedup |
|---|---|---|:--:|:--:|:--:|
| **cameraman** `Gray{N0f8}` 512² | down 4× | 128² | 72.5 µs | 41.2 µs | **1.76×** |
| | down 2× | 256² | 288.5 µs | 158.9 µs | **1.82×** |
| | up 2× | 1024² | 5.79 ms | 1.64 ms | **3.53×** |
| | up 4× | 2048² | 23.61 ms | 5.67 ms | **4.16×** |
| **earth_apollo17** `RGB{N0f8}` 9 MP | down 4× | 751×750 | 6.97 ms | 3.79 ms | **1.84×** |
| | down 2× | 1501×1500 | 27.71 ms | 14.91 ms | **1.86×** |
| | up 2× | 6004×6000 | 542.0 ms | 137.2 ms | **3.95×** |
| **mri** `Gray{N0f8}` 3-D 226×186×27 | down 4× | 57×47×7 | 142.7 µs | 75.1 µs | **1.90×** |
| | down 2× | 113×93×14 | 1.14 ms | 585.8 µs | **1.94×** |
| | up 2× | 452×372×54 | 100.7 ms | 23.33 ms | **4.32×** |
| | up 4× | 904×744×108 | 799.4 ms | 151.7 ms | **5.27×** |

This needs FastInterpolations 0.4.16, where the API it uses became public.

Thanks for taking a look!


mgyoo86:
I just realized that `ImageTransformations` supports from julia 1.6,
whereas `FastInterpolations` supports from julia 1.10 (LTS).

I've left this PR as-is for now to keep it simple to review. 
If you think it's worth merging but bumping the minimum Julia version is too much, I can instead make FastInterpolations a weak dependency, but its accessibility to users would be significnalty reduced.

Please let me know what you think, if you're intersted in this PR.
Thank you!

